PR #173, opened 2017-02-02
Adding Mapping Validator in the Mapping Dialog Panel

sinhaaman:
Project- Semantic Technology


sinhaaman:
About commit ID 4015ce4.

Earlier the SPARQL Query passed was:

CONSTRUCT {?s ?p ?o} WHERE {?s ?p ?o}.

Now, the construct query is revised and made by parsing the target query.

Issue #211, opened 2017-10-02
Duplicate attribute name due to error parsing SQL


fanavarro:
Hi, I'm using ontop in my project and I have decided to update it from 1.18 to 3.0.0-beta-1 version. I downloaded the protege bundle in order to try the new version but it threw an error (duplicate attribute name) when I tried to start the reasoner. The ontology and the mapping files works on the previous version. 

I have downloaded the API examples and I have modified one of them in order to use my ontology and mappings and to be able to debug the code. I've figured out that the bug is in the function extract of the class SelectQueryAttributeExtractor. The conflictive mapping has the following SQL query:

```
SELECT ALMAES001.IDART,
     ALMAES001.CODART,
     ALMAES001.IDFAMI,
     ALMAES001.DESCRIP,
     ALMAES001.DESCRII,
     ALMAES001.MEDIDA,
     ALMAES001.DESCUENTO,
     ALMAES001.UNIDADES,
     ALMAES001.UDSENV,
     ALMAES001.IDTIPIVA,
     ALMAES001.UPC,
     ALMAES001.PVP1, ALMAES001.PVP2, ALMAES001.PVP3, ALMAES001.PVP4, ALMAES001.PVP5,
     to_char(ALMAES001.FECALTA,'YYYY-MM-DD') AS FECALTA,
     to_char(ALMAES001.FECBLO,'YYYY-MM-DD') AS FECBLO,
     ALMAES001.STCMIN,
     ALMAES001.STCMAX,
     ALMAES001.PUNRUP,
     ALMAES001.UBICACION,
     ALMAES001.IDCTACOMPRA,
     ALMAES001.IDCTAVENTA,
     ALESFO001.PRECIOFINAL
	             
FROM ALMAES001
LEFT JOIN ALESFO001 ON ALMAES001.IDART = ALESFO001.IDART
```
It has a left join clause, which are not supported. Because of this, the method "extract" catch the exception and try to parse the SQL in order to retrieve the name of the columns:
```

    public ImmutableList<QuotedID> extract(String sql) throws InvalidSelectQueryException {

        try {
            ImmutableMap<QualifiedAttributeID, Variable> attrs = sqae.parse(sql).getAttributes();

            return attrs.keySet().stream()
                    .filter(id -> id.getRelation() == null)
                    .map(id -> id.getAttribute())
                    .collect(ImmutableCollectors.toList());
        }
        catch (Exception e) {
[...]

PR #212, opened 2017-10-04
Fixed SQL parsing error when functions appear on select clause.

fanavarro:
Related to #211.
There was a bug parsing SQL. Ontop was expecting that identifiers in SELECT clause were separated by commas in order to extract the column names. This produces an error if a function appears. In my case
`to_char(ALMAES001.FECALTA,'YYYY-MM-DD') AS FECALTA,`
was splitted into 
`to_char(ALMAES001.FECALTA`
and
`'YYYY-MM-DD') AS FECALTA`

Each string is treated individually so the application tried to insert FECALTA twice, which causes a "Duplicate attribute name" error.

What I've done to avoid this behaviour is to detect functions with the regexp 
`\w+\s*\(.*\)\s*(as|AS)`
and to delete it, keeping the column name given to the result of the function.

The use of "AS" clause is mandatory to avoid the bug.


skomlaebri:
Thanks for reporting the problem we solved the issue differently and refactored part of the code.

Issue #242, opened 2018-01-29
Language tag appended in literal part of a lang string.


fanavarro:
Hi all,

I'm seeing a strange behavior when I query about literals with a language tag established. Ontop returns an OWLLiteralImplPlain with the property "lang" filled correctly, but this language tag is appended to the literal value too. For example, the following query about the rdfs:comment of an ontology class

```
SELECT ?instancia ?comment  WHERE {
	VALUES ?instancia { <http://www.basecia.es/ontologia#CuentaContableActivos> } .
	?instancia <http://www.w3.org/2000/01/rdf-schema#comment> ?comment .
}
```
returns a OWLLiteralImplPlain with the following properties:

- lang = "es"

- literal = "Cuenta bancaria interna de activos.es"

As you can see, the language tag "es" is appended to the literal part.

I'm using the code from the main branch of git repository.
I'm going to do some debug. I'll write if I found something.

The files used (ontology, mappings and properties) are attached.

Cheers,
Fran.

[obda.zip](https://github.com/ontop/ontop/files/1673140/obda.zip)

fanavarro:
I've found that a lang string is represented as a function. In my case, the following one:
`http://www.w3.org/1999/02/22-rdf-syntax-ns#langString("Cuenta bancaria interna de activos.","es")`
This function has 2 terms: the literal part ("Cuenta bancaria interna de activos.") and the language tag ("es").

The method "getSQLString(Term term, AliasIndex index, boolean useBrackets)" from OneShotSQLGeneratorEngine detects the type of the function as data type function. After that, getSQLStringForTemplateFunction is called. In this function, the SQL of the terms are concatenated so, finally, we have the following SQL line:

`('Cuenta bancaria interna de activos.' || REPLACE(REPLACE(REPLACE(REPLACE(REPLACE(REPLACE(REPLACE(REPLACE(REPLACE(REPLACE(REPLACE(REPLACE(REPLACE(REPLACE(REPLACE(REPLACE(REPLACE(REPLACE(REPLACE(CAST('es' AS VARCHAR(4000)), ' ', '%20'), '!', '%21'), '''', '%22'), '#', '%23'), '$', '%24'), '&', '%26'), '(', '%28'), ')', '%29'), '*', '%2A'), '+', '%2B'), ',', '%2C'), '/', '%2F'), ':', '%3A'), ';', '%3B'), '=', '%3D'), '?', '%3F'), '@', '%40'), '[', '%5B'), ']', '%5D'))`

[...]

PR #243, opened 2018-01-29
Check langString functions (#242)

fanavarro:
This is a fix for #242 so there is more info in the issue page.
I've added a check for langString functions when SQL is going to be built.

Before this fix, when langString function was detected the method "getSQLStringForTemplateFunction" was called. This method takes the terms of the function (in this case literal part and language tag) and concatenates them causing a response of type "literal parten"@en (language tag appended to the literal part).

Now, when a langString function is detected, only the first argument is treated.


juliencorman:
Thank you Francisco for the fix!

Cheers,
Julien

PR #265, opened 2018-05-30
Limit on adding EmptyNodes

ghuron:
When dealing with thousands of mappings, unfolding even simple queries leads to addition of large number of EmptyNodes (in mergeSubQuery()) and their subsequent removal (in apply()). Artificial limitation of them by 2 per parent speed up unfolding at least 3 times in my environment.

### DIFF
--- a/core/optimization/src/main/java/it/unibz/inf/ontop/iq/executor/merging/QueryMergingExecutorImpl.java
+++ b/core/optimization/src/main/java/it/unibz/inf/ontop/iq/executor/merging/QueryMergingExecutorImpl.java
@@ -267,19 +267,21 @@ public class QueryMergingExecutorImpl implements QueryMergingExecutor {
             QueryNode nodeToInsert = getNodeToInsert(transformation, treeComponent);
 
             /**
-             * Adds the transformed node to the tree.
+             * Adds the transformed node to the tree if its not empty or there are less then 2 nodes for parent 
              */
-            treeComponent.addChild(transformation.getTransformedParent(),
-                    nodeToInsert,
-                    transformation.getOptionalPosition(), false);
-
-            Optional<? extends ImmutableSubstitution<? extends ImmutableTerm>> substitutionToPropagate =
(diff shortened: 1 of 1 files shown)


ghuron:
I've signed the agreement and send scan. Let me briefly describe what we are trying to do.
One of the projects behind wikipedia called wikidata. It is basically graph database, that contains some information about things, described in wikipedias alongside with their relations. Primary query language there is SPARQL, and they utilize blazegraph for that. Despite the fact that wikidata is closely related to wikipedia, not every applicable statement from wikipedia articles are migrated to wikidata yet. In order to simplify migration process, a colleague of mine created simple sparql query endpoint based on ontop, that can be utilized via federated query (see https://github.com/Tpt/mw2sparql). The ontology is relatively simple (https://github.com/Tpt/mw2sparql/blob/master/src/main/resources/ontology.ttl), but the problem is that ideally we want it for 880 project databases (e.g. en.wikipedia.org, de.wikipedia.org, etc). All databases shares exactly the same structure, so we basically apply https://github.com/Tpt/mw2sparql/blob/master/src/main/resources/mapping.ttl 880 times with different parameters. Major performance problem lies within BasicUnfolder, my hack above only slightly improve situation, so I've created my own unfolder (https://github.com/Tpt/mw2sparql/blob/master/src/main/java/org/mediawiki/sparql/mwontop/utils/SiteSpecificUnfolder.java) with some hacky tricks, that applies to most frequently used queries. I've connected it using it.unibz.inf.ontop.answering.reformulation.unfolding. QueryUnfolder property. It does improve situation dramatically, but not on all queries. I'm looking forward to your beta3, or if you want me to test any specific branch of ontop in terms of performance - I'm ready to help.
There is one more thing, that dramatically reduce performance in my setup. Federated queries extensively used VALUES sparql keyword (sometimes with hundreds or thousands items), which is currently translated into sql joins. If we can translated it into IN SQL keyword, it would greatly improve db query performance

PR #272, opened 2018-10-08
version3

JensPiegsa:
Building Ontop on a machine with German Locale led to failing unit test. This pull request contains a small fix.


bcogrel:
Sorry, unfortunately, we didn't take the time to include it into 3.0. 

The way RDF constants are built has been refactored in 4.0, causing this class not to exist anymore.

PR #326, opened 2020-02-17
Implemented streaming on Sparql endpoint.

guillemrullfort:
This pull request modifies the SPARQL endpoint so that it streams the query result to the client rather than keeping it entirely in memory.


bcogrel:
Thanks a lot Guillem! It makes a lot of sense.

PR #364, opened 2020-11-10
Javac-11-.0.9 RH says root childeren generic is off.

JervenBolleman:
This is a small change in the generic here. The new code compiles both with java8 and java11 while the old one seems to fail in java11.


bcogrel:
Thanks!

PR #370, opened 2020-11-27
Added memory improvements and introduced streaming for construct/export operations

tokovach:
Added memory improvements and introduced streaming for construct queries

- Simplified Binding set structure to use arrays with bindings rather than expensive structures such as LinkedHashMaps

- Updated DefaultSimpleGraphResultSet to implement an iterator that directly returns statements rather than the more expensive implementation RDFFact.

- Updated OntopRepositoryConnection not to re-iterate over graph result set but rather to stream it as a closable one.

- Updated OntopGraphQuery and QuestStatement, as well as DefaultSimpleGraphResultSet to stream directly statements rather than first store them directly in memory, which can be quite expensive when the number of triples is more than 1 million.


bcogrel:
Hi Tomas,

Thanks a lot for your PR.

In general, I think it is going in the right direction. Apparently, it leads to a substantial reduction of the memory usage for large result sets.

It changes several interfaces and in particular:
 1. `GraphResultSet`, which is now extending `Iterable<Statement>` . This adds an additional dependency to a RDF4J interface in a core module (`ontop-reformulation-core`), but I think it is the price to pay for avoiding to create 2 objects for the same output triple/quad.
2. `OntopBindingSet` is now using arrays instead of `ImmutableList`s.  I think that's fine, as it is a low-level interface.

At first glance, by looking at the code, bnode anonymisation (based on a salt) seems to be preserved. I will need to double-check it.

I noticed that most of changes in the current state of this PR are related to code formatting, which makes reviewing harder and the `git blame` command less powerful. 

Won't you mind removing the code formatting changes so as to keep this PR focused on the important changes?

Best,
Benjamin

tokovach:
Hi Benjamin,

Thank you for the feedback. Getting right to it!

Kind regards,
Tomas

tokovach:
Hi Benjamin,

I have added further changes that make sure the DefaultSimpleGraphResultSet conforms to the same logic before ( for close() method) so here are the changes:

- Updated DefaultSimpleGraphResultSet's iterator to implement ClosableIteration as it has a close() method used by RDF4J to close connection to result set. This also helps with conformance to GraphResultSet interface as now connection to Tuple Result Set can be closed without having to iterate entire result set. Thus, making sure that no connection is left hanging.
 
 - Updated OntopGraphQuery to return directly the GraphResultSet's iterator as it is now closable, rather than to wrap it in CloseableIteratorIteration which has an empty close() method.
 
 - Introduced a new exception that is needed in order to implement a closeable iteration.

Hope this is useful!

Kind regards,
Tomas

bcogrel:
Hi Tomas,

Sorry for not replying earlier but I had a busy week.

The new changes you have pushed are interesting, but I think now there are too many changes for a single PR. In particular it is conflicting with the current modular structure of Ontop. I think we need to go step by step. Let me explain a bit the context.

Core modules like `ontop-reformulation-core` and `ontop-system-core` were so far expected to be "independent" from binding libraries like RDF4J, Jena and OWLAPI. By independent, we mainly meant that their interfaces should not appear in our core interfaces. In that line, we are for instance using Apache Commons RDF to avoid having IRIs tightly bound to RDF4J. Also the fact that RDF4J appears in `ontop-reformulation-core` to implement `InputQueryTranslator` is an implementation detail, which is for instance made invisible to a class like `QuestQueryProcessor`. 

Your proposal is challenging this design in several ways:
  1. `GraphResultSet` would be made dependent on `Statement` (already mentioned in my previous comment)
  2.  In your last commit `GraphResultSet` would also depend on the `CloseableIteration` RDF4J interface for better resource management than with regular iterators.
  3.  Core modules like `ontop-system-core` may directly throw RDF4J exceptions (unchecked) instead of `OntopQueryEngineException`s (checked). `QueryEvaluationException`  (from RDF4J) even appears in the `GraphResultSet` interface (from  `ontop-reformulation-core`).

I view the point 3 as problematic as it makes it very unclear what to expect from these core modules. My impression is that it could be avoided with no major effort so I would suggest to stick to Ontop exceptions for the moment.

Points 1 and 2 are more open and invite us to revisit the relationship between Ontop and RDF4J. It is true that over time RDF4J has become the dominant binding API for Ontop. Offering bindings for another major library does not look to me very attractive anymore, given the availability of the HTTP interface of the SPARQL endpoint and the growing popularity of micro-service architectures. We will discuss it next week with the other developers of the project.

In the meantime, I would suggest the following:
  - Take the 2 first commits, make sure code reformatting is kept minimal and that no RDF4 exceptions are thrown instead of `OntopQueryEngineException`s  (I could see for instance several `SailException`-s), and squash them into one commit.
  - Move the last commit (about `CloseableIteration`) into another branch.

If you want, we could have a chat next week.

Best,
Benjamin

tokovach:
Hi Benjamin,

Thank you for your feedback. I completely agree with you and will move the latter commit in a new merge request as well as fix the formatting for the first two commits.

Would be very happy to have a chat next week regarding the commit and idea behind the project architecture. Could you please drop a line with convenient time for you next week to tomas.kovachev@ontotext.com.

Kind regards and happy holidays!
Tomas

tokovach:
Hi Benjamin,

Following our discussion I have cleared out the unnecessary RDF4J dependencies from core modules and have created Ontop interfaces for the closeable iterations. Here is a summary of the changes:

- Updated OntopGraphQuery and DefaultSimpleGraphResultSet to stream RDFFacts rather than RDF4J's statements

- Removed dependencies on RDF4J from core modules and introduced Closeable Iterator interfaces for streaming results

- Added RDFFactCloseableIterator abstract class that can be used for streaming results to RDF Handlers and supports close() for connection, where you only have to override handleClose() to allow for connection closure

- Moved back RDF4JHelper to rdf4j module and added a wrapping OntopCloseableStatementIteration that translates RDFFact to RDF4J's Statement and allows for closing of underlying connection to db

Hope theses changes are useful!

Kind regards,
Tomas

engine/reformulation/core/src/main/java/it/unibz/inf/ontop/answering/resultset/SimpleGraphResultSet.java
```diff
@@ -1,17 +1,19 @@
...
 
     /**
      * TODO: remove this hack
      */
-	@Deprecated
-    void addNewResult(RDFFact assertion);
+    @Deprecated
+    void addNewResult(RDFFact statement);
 
+    void addStatementClosable(AutoCloseable sqlStatement);
```
bcogrel, line 18: From what I have seen the statement is not a SQL statement but an OntopStatement. 

What about removing this method from the interface, and passing the OntopStatement as an argument when building the SimpleGraphResultSet?

binding/rdf4j/src/main/java/it/unibz/inf/ontop/rdf4j/query/impl/OntopGraphQuery.java
```diff
@@ -39,52 +33,28 @@ public OntopGraphQuery(String queryString, ParsedQuery parsedQuery, String baseI
...
-			
+	    GraphSPARQLQuery query =
+    	    isConstruct
+        	    ? inputQueryFactory.createConstructQuery(getQueryString(), parsedQuery, bindings)
+            	: inputQueryFactory.createDescribeQuery(getQueryString(), parsedQuery, bindings);
+		try
+		{
+			OntopStatement stm = conn.createStatement();
+			SimpleGraphResultSet res = stm.execute(query, getHttpHeaders());
+			res.addStatementClosable(stm);
```
bcogrel, line 50: The last line could be removed after refactoring (see other comment)

engine/reformulation/core/src/main/java/it/unibz/inf/ontop/answering/resultset/GraphResultSet.java
```diff
@@ -1,10 +1,16 @@
...
 import it.unibz.inf.ontop.exception.OntopQueryAnsweringException;
 import it.unibz.inf.ontop.spec.ontology.RDFFact;
 
 public interface GraphResultSet<X extends OntopQueryAnsweringException> extends IterativeOBDAResultSet<RDFFact, X> {
 
     @Override
-    RDFFact next() throws X;
+    RDFFact next() throws X, OntopConnectionException;
+
+    RDFFactCloseableIterator iterator();
```
bcogrel, line 14: `RDFFactCloseableIterator` is not an interface. Please use `OntopCloseableIterator<RDFFact, OntopConnectionException>`

engine/reformulation/core/src/main/java/it/unibz/inf/ontop/answering/resultset/OntopCloseableIterator.java
```diff
@@ -0,0 +1,15 @@
...
+import it.unibz.inf.ontop.spec.ontology.RDFFact;
+
+public interface OntopCloseableIterator<X extends RDFFact, Y extends OntopConnectionException> extends AutoCloseable{
+
+    boolean hasNext() throws OntopConnectionException, OntopResultConversionException;
+
+    X next() throws Y;
+
+    void remove() throws Y;
+
```
bcogrel, line 14:
```java
    @Override
    void close() throws Y;
```
is missing

engine/reformulation/core/src/main/java/it/unibz/inf/ontop/answering/resultset/SimpleGraphResultSet.java
bcogrel, line 18: From what I have seen the statement is not a SQL statement but an `OntopStatement` (misleading name).

What about removing this method from the interface, and passing the `OntopStatement` as an argument when building the `SimpleGraphResultSet`?

engine/system/core/src/main/java/it/unibz/inf/ontop/answering/connection/impl/QuestStatement.java
```diff
@@ -224,8 +227,10 @@ private SimpleGraphResultSet executeDescribeQuery(DescribeQuery inputQuery, Immu
 					describeResultSet = set;
 				} else if (set != null) {
 					// 2nd and manyth times execute, but collect result into one object
-					while (set.hasNext())
-						describeResultSet.addNewResult(set.next());
+					RDFFactCloseableIterator iterator = set.iterator();
```
bcogrel, line 230: Will have to use the interface `OntopCloseableIterator<RDFFact, OntopConnectionException>`

engine/system/core/src/main/java/it/unibz/inf/ontop/answering/connection/impl/QuestStatement.java
```diff
@@ -238,8 +243,10 @@ private SimpleGraphResultSet executeDescribeQuery(DescribeQuery inputQuery, Immu
 				if (describeResultSet == null) { // just for the first time
 					describeResultSet = set;
 				} else if (set != null) {
-					while (set.hasNext())
-						describeResultSet.addNewResult(set.next());
+					RDFFactCloseableIterator iterator = set.iterator();
```
bcogrel, line 246: Will have to use the interface `OntopCloseableIterator<RDFFact, OntopConnectionException>`

engine/system/core/src/main/java/it/unibz/inf/ontop/materialization/impl/DefaultMaterializedGraphResultSet.java
```diff
@@ -182,6 +183,12 @@ public RDFFact next() throws OntopQueryAnsweringException {
         return null;
     }
 
+    //TODO implement a closable iterator
+    @Override
+    public RDFFactCloseableIterator iterator() {
```
bcogrel, line 188: Please use the interface `OntopCloseableIterator<RDFFact, OntopConnectionException>` instead

bcogrel (changes requested):
Hi Tomas,

Great move! 

I think we are almost there. Could you please have a look at the small changes I suggested?
After that, in my view, we should be good to go.

Thanks for the good work!
Benjamin

engine/reformulation/core/src/main/java/it/unibz/inf/ontop/answering/resultset/SimpleGraphResultSet.java
tokovach: I think this might also be passed as a java.sql Statement to the DefaultSimpleGraphResultSet from inside the SqlQuestStatement where the sqlStatement is held

tokovach:
Hi Benjamin,

Following your comments, I have updated GraphResultSet iterator method to return OntopCloseableIterator interface and also the DefaultSimpleGraphResultSet to pass java.sql.Statement to the iterator and removed unnecessary method addClosableStatement() from SimpleGraphResultSet

bcogrel:
Hi Tomas,

Great, thanks! We are now good to go.

I just did a small change: `DefaultSimpleGraphResultSet` now receives an `OntopStatement` instead of a SQL statement, as it is in a module (`ontop-system-core`) which is independent from SQL. `OntopStatement` is taking care itself of closing the underlying low-level DB statement.

I am about push everything into the main branch.

Thanks again for the good work,
Benjamin

PR #393, opened 2021-02-17
POM and assembly restructuring to replace build-release.sh script

fracorco:
This pull request contains modified POM files and assembly descriptors so that the following release files:

* `it.unibz.inf.ontop.protege-4.1.0-beta-1-SNAPSHOT.jar`
* `ontop-cli-4.1.0-beta-1-SNAPSHOT.zip`
* `ontop-jetty-bundle-4.1.0-beta-1-SNAPSHOT.zip`
* `ontop-protege-bundle-4.1.0-beta-1-SNAPSHOT.zip`
* `ontop-tomcat-bundle-4.1.0-beta-1-SNAPSHOT.zip`
* `ontop-webapps-4.1.0-beta-1-SNAPSHOT.zip`

which are currently generated by the `build-release.sh` script, can be created entirely by maven by enabling a `release` profile:

```
  mvn package -DskipTests -Prelease
```

The download of Jetty, Tomcat and Protege and their repackaging with ontop binaries is done via a mix of maven-dependency-plugin, download-maven-plugin and the maven-assembly-plugin, and all the release files appear under `build/distribution/target`, so `mvn clean` will wipe them out, as expected.

This refactoring has been tested using JDK 8 on Linux and Windows, and it generates the same release files of the original build script (pending changes in JAR files due to compile process not consistently producing the same bytecode).

**Important Note**: This setup adopts an all or nothing approach, whereby the user can either generate all the release artifacts (the files listed above) by supplying `-Prelease`, or none of them. As a user (and developer) I have been mostly interested in the `ontop-cli` artifact and I would find inconvenient to spend bandwidth (for download) and time to generate the other artifacts when I'm not interested in them. If you find it useful, I may investigate the use of finer-grained build options / profiles to select each artifact independently, and `release` to enable all of them.


fracorco:
Here are some other changes related to POM files you may consider:
* extend main `pom.xml` section `dependencyManagement` with dependencies for project modules, so not to have to declare their version every time they are imported;
* extend main `pom.xml` section `dependencyManagement`  with dependencies that currently appear only in modules, so to centralize dependency version declaration (via properties);
* evaluate whether to move all additional properties declarations scattered in child modules to main `pom.xml` (may be good to centralize dependency versions, maybe not desired for some module-specific properties)
* clean up duplicate (i.e., same value) property declarations in child modules, e.g., `project.build.sourceEncoding = UTF-8` in `ontop-owlapi` and `ontop-optimization`
* clean up redundant plugin declarations in child modules, e.g., `maven-jar-plugin` in `ontop-optimization` and `ontop-obda-core`
* to control per-module deployment, instead of adding a `build/plugins/plugin` block to each affected module, define a boolean property like `deploy.skip` in main `pom.xml`, refer to it in main `pom.xml` deploy plugin configuration, and override the property in the modules where it is needed
  (similar solution may possibly work for specifying main class of a module)

  In `main pom.xml`:
  
  ```
    <properties>
      <deploy.skip>false</deploy.skip>
    </properties>
    
    <build>
        <plugins>
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-deploy-plugin</artifactId>
                <configuration>
                    <skip>${deploy.skip}</skip>
                </configuration>
            </plugin>
        </plugins>
    </build>
  ```
  In childe module `pom.xml` where to disable deployment:
  
  ```
  <properties>
    <deploy.skip>true</deploy.skip>   
  </properties>
  ```

* may configure maven-dependency-plugin goal `analyze-only` to verify dependencies (e.g., that no unused dependency is imported and all used dependencies are imported) as part of build, either by default, or at release time only (e.g., with `-Prelease`)

* may configure [git-commit-id-maven-plugin](https://github.com/git-commit-id/git-commit-id-maven-plugin) to include GIT metadata (e.g., commit ID) in generated JAR files, which may be useful for testing

* similarly, may configure `build-info` goal of [spring-boot-maven-plugin](https://docs.spring.io/spring-boot/docs/current/maven-plugin/reference/htmlsingle/#goals-build-info) to generate `build-info.properties` metadata to be included in JAR files (work on any module, not just spring-related ones)

* may configure [license-maven-plugin](https://www.mojohaus.org/license-maven-plugin/aggregate-add-third-party-mojo.html) in main `pom.xml` to better merge similar licenses (e.g., different spellings of same license name) when automatically generating `THIRD-PARTY.txt` file listing licenses of included dependencies

ghxiao:
Hi @fracorco, all of your suggestions are very nice. I have turned them into a list of actions and will add a few more. Could you take care of them?

fracorco:
> Hi @fracorco, all of your suggestions are very nice. I have turned them into a list of actions and will add a few more. Could you take care of them?

Sure. As I'll likely have to change most of POM files, in the meanwhile let me know if there are other desiderata or requirements you'd like to take into account.

bcogrel:
Hi @fracorco , 

I agree, it would be nice to be able to generate only the release file we are interested in. In particular, when we only need `ontop-cli-4.1.0-beta-1-SNAPSHOT.zip` for building the Docker image (separated script).

Also, I think we can remove  `ontop-jetty-bundle-4.1.0-beta-1-SNAPSHOT.zip` and `ontop-tomcat-bundle-4.1.0-beta-1-SNAPSHOT.zip` as we don't want to take care of them anymore.  We may still allow people to build `ontop-webapps-4.1.0-beta-1-SNAPSHOT.zip` on demand, but it would be better not having it part on the `release` profile.

I was also wondering if we could stop publishing the following modules on Maven Central:
- `ontop-test`
- `ontop-docker-tests`
- `ontop-rdb2rdf-compliance`
- `ontop-semantic-index`
- `ontop-sparql-compliance`
- `ontop-build`
- `ontop-distribution`

Best,
Benjamin

fracorco:
Hi @bcogrel, @ghxiao,

summing up, we want the following profiles / build properties:
* `cli`, to generate `ontop-cli-4.1.0-beta-1-SNAPSHOT.zip`
* `release`, to generate both cli and protege artifacts, i.e., 
   - `ontop-cli-4.1.0-beta-1-SNAPSHOT.zip`
   - `it.unibz.inf.ontop.protege-4.1.0-beta-1-SNAPSHOT.jar`
   - `ontop-protege-bundle-platform-independent-4.1.0-beta-1-SNAPSHOT.zip`
   - `ontop-protege-bundle-win-4.1.0-beta-1-SNAPSHOT.zip`
   - `ontop-protege-bundle-linux-4.1.0-beta-1-SNAPSHOT.tar.gz`
   - `ontop-protege-bundle-os-x-4.1.0-beta-1-SNAPSHOT.zip`
* `webapps`, to generate `ontop-webapps-4.1.0-beta-1-SNAPSHOT.zip`

The artifacts produced by `release` ends up in GitHub releases. The jetty and tomcat bundles are discarded. I'll set this up.

Concerning what to publish on Maven Central (provided this will be configurable with a per-module property in the POMs), my 2 cents are that any test JAR that may be reused by users extending Ontop (e.g., because it provides some common test code or resources to build on) is worth being published, as then it can be fully / selectively imported in users' test code. I did this with the compliance test suites of RDF4J to test an extended SPARQL parser.

Besides, I'm editing the POM files and using a couple of auto-formatting plugins to ensure consistent structure and indentation of those files (which is not the case now) - to be run on-demand now & in the future via something like `mvn process-sources -Dautoformat`. While this eases the task for me, a drawback is that it will be difficult to track what changed in POM files after merging this branch (changes will be a lot anyway). If there are objections let me know and I'll resort to a more conservative manual editing approach.

bcogrel:
Hi @fracorco ,

Ok I see, if you think it is useful, we can keep them.

Regarding auto-formatting, it is probably ok for Maven, as I don't think that navigating in the history for these kind of files is so important. 

For code base of Ontop, we currently follow a conservative approach, as git blame/annotate command is very valuable to navigate across all these years of coding.

fracorco:
Hi @bcogrel,

> Ok I see, if you think it is useful, we can keep them.

I'm not familiar with some of those modules you mentioned so whether they can be used outside the project by people integrating / extending Ontop is something I cannot judge well. Maybe (at minimum) `ontop-build` and `ontop-distribution` can be safely not deployed to Maven central. As a reference, I checked RDF4J and there they have a [`rdf4j-assembly` module](https://github.com/eclipse/rdf4j/tree/master/assembly) (the one generating the SDK and onejar artifacts) they stopped publishing on Maven central starting with RDF4J version 3 (see [here](https://repo1.maven.org/maven2/org/eclipse/rdf4j/rdf4j-assembly/)). In this specific case, this is done not by skipping the deploy plugin but by not listing the module as a child of the root module, which means they have to specifically run Maven on it vs running it as part of an invocation on the root module (all other things being equal, I would choose skipping the deploy plugin).

> Regarding auto-formatting, it is probably ok for Maven, as I don't think that navigating in the history for these kind of files is so important.

Perfect, thanks.

> For code base of Ontop, we currently follow a conservative approach, as git blame/annotate command is very valuable to navigate across all these years of coding.

I understand and good to know. I wasn't considering re-formatting existing Java code, although auto-formatting when coding is something that I definitely use. You might still want to document Java coding convention for new code contributions (i.e., new modules, new source code files).

PR #410, opened 2021-03-02
Feature/GitHub maven repo

fracorco:
This pull request aims at replacing custom Maven repositories `clojars` and `bolzano-nexus-public`, which are deemed too slow, with a new custom Maven repository hosted directly on GitHub, and consisting of a [Git repository](https://github.com/ontop/ontop-maven-repo) owned by the ontop GitHub team. 

Everything works, but I invite you to consider the following:

1. Decide on repository name `ontop-maven-repo`. Previously we discussed using `third-party-jars`, but I realized that this repository may be also used to deploy Ontop-related artifacts (e.g., Ontop snapshots not ready for Maven central, or the JARs of some other project under the ontop team that is not meant/ready to go on Maven central), so I switched to a more general name.

2. Read the [README.txt](https://github.com/ontop/ontop-maven-repo/blob/master/README.md) of such repository, focusing in particular on the disclaimers about license and third-party use (the instructions have been tested and work fine). Note that if we pick a different repository name the content of the README file should be adapted correspondingly.

3. Please be aware that this feature branch also include a somehow unrelated change, i.e., the pinning of `tomcat-embed-core` to a specific version in the root `pom.xml`. That was needed for some reason in order to satisfy the `maven-enforcer-plugin`. If you prefer, we can implement this change as a separate bug fix.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -182,6 +182,7 @@
         <semargl-sesame.version>0.6.1</semargl-sesame.version>
         <semargl.version>0.7</semargl.version>
         <sesame.version>2.7.16</sesame.version>
+        <tomcat-embed-core.version>9.0.37</tomcat-embed-core.version>
     </properties>
 
     <dependencyManagement>
@@ -1062,6 +1063,11 @@
                 <groupId>org.junit.jupiter</groupId>
                 <artifactId>junit-jupiter-api</artifactId>
                 <version>${junit-jupiter-api.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.tomcat.embed</groupId>
+                <artifactId>tomcat-embed-core</artifactId>
+                <version>${tomcat-embed-core.version}</version>
             </dependency>
         </dependencies>
     </dependencyManagement>
--- a/test/docker-tests/pom.xml
+++ b/test/docker-tests/pom.xml
@@ -143,8 +143,8 @@
             <snapshots>
                 <enabled>false</enabled>
             </snapshots>
-            <id>third-party-jars</id>
-            <url>https://raw.githubusercontent.com/fracorco/third-party-jars/master/releases/</url>
+            <id>ontop-maven-repo</id>
+            <url>https://raw.githubusercontent.com/ontop/ontop-maven-repo/master/repository/</url>
         </repository>
     </repositories>
 
--- a/test/docker-tests/pom.xml
+++ b/test/docker-tests/pom.xml
@@ -131,26 +131,20 @@
 
     <repositories>
         <repository>
-            <!-- for 3rd party JDBC drivers -->
-            <releases>
-                <enabled>true</enabled>
-            </releases>
-            <snapshots>
-                <enabled>true</enabled>
-            </snapshots>
-            <id>bolzano-nexus-public</id>
-            <url>http://obdavm.inf.unibz.it:8080/nexus/content/groups/public/</url>
-        </repository>
-        <repository>
-            <!-- for monetdb -->
+            <!--
+                For 3rd party JDBC drivers:
+                - madgit.adp:adp-jdbc:0.1.2
+                - monetdb:monetdb-jdbc:11.19.15
+                - com.dremio.jdbc:dremio-jdbc:1.4.4
+            -->
             <releases>
                 <enabled>true</enabled>
             </releases>
             <snapshots>
                 <enabled>false</enabled>
             </snapshots>
-            <id>clojars.org</id>
-            <url>http://clojars.org/repo</url>
+            <id>third-party-jars</id>
+            <url>https://raw.githubusercontent.com/fracorco/third-party-jars/master/releases/</url>
         </repository>
     </repositories>
 
--- a/test/docker-tests/pom.xml
+++ b/test/docker-tests/pom.xml
@@ -143,8 +143,8 @@
             <snapshots>
                 <enabled>false</enabled>
             </snapshots>
-            <id>ontop-maven-repo</id>
-            <url>https://raw.githubusercontent.com/ontop/ontop-maven-repo/master/repository/</url>
+            <id>ontop-3rd-party-maven-repo</id>
+            <url>https://raw.githubusercontent.com/ontop/ontop-3rd-party-maven-repo/master/repository/</url>
         </repository>
     </repositories>
 


fracorco:
For future reference, we document here an alternative option (discussed via email and currently discarded) to create a GitHub-hosted Maven repository, based on [GitHub Packages](https://docs.github.com/en/packages).

GitHub Packages is a service that can be specifically used for deploying binary packages [such as Maven ones](https://docs.github.com/en/packages/guides/configuring-apache-maven-for-use-with-github-packages#installing-a-package) (and also npm, dotnet, ruby, docker). Basically, associated to each GitHub repository (like `ontop/ontop`) there's a GitHub Packages instance (url: `https://maven.pkg.github.com/ontop/ontop`) that may act as a Maven repository. The service is available also for free public repositories and appear to work

Two issues so far:

1. Token-based authentication is needed both to upload and download a package (i.e., dependency jar), where the token is something that each GitHub user can [generate](https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token) for himself (GitHub Settings > Developer settings > Personal access tokens > Generate new token).

   (1a) for uploading, we need that every user with write access to ontop may be able to upload a jar, which is something that need further testing using multiple GitHub accounts
   
    (1b) for downloading, we need any anonymous user to be able to clone and compile ontop, which is currently [unsupported](https://github.community/t/how-to-allow-unauthorised-read-access-to-github-packages-maven-repository/115517/4). GitHub plans to support anonymous read access later this autumn. In the meanwhile we would need to implement the hack [here](https://github.com/jcansdale-test/maven-consume), which basically consists in creating a personal token with read package access as only permission to share with the world.

2. The deployed third-party jars are listed as project "Packages" in the project home page on GitHub (right column, at the bottom). Not strictly a problem but may still confuse users (why `monetdb.monetdb-jdbc` is a "package" of Ontop?).

Summing up, the GitHub Packages option might be more appropriate but its current implementation appears immature, especially concerning the workaround needed for (1b) which is pretty invasive and may cause some issue when running Maven not on the root directory but in a module (as .mvn/... files used in the hack are not picked up automatically, which means further command line options must be given to pick them). Besides, also (1a) requires further Maven setup by each Ontop developer wishing to upload a JAR. For these reasons, we are documenting but otherwise discarding this option.

Should GitHub Packages being considered in the future, the steps for using it are the following:

1. each Ontop developer wishing to upload a JAR need to create a personal GitHub token with at least permission `write:packages`; then, he/she need to edit file `~/.m2/settings.xml` adding inside `<servers>`
   ```xml
   <server>
     <id>github</id>
     <username>YOUR_GITHUB_USERNAME</username>
     <password>YOUR_GITHUB_TOKEN</password>
   </server>
   ```

2. deploy the required JARs to GitHub Packages via Maven command:
   ```
   mvn deploy:deploy-file \
       -DrepositoryId=github \
       -Durl=https://maven.pkg.github.com/ontop/ontop \
       -Dfile=PATH_TO_FILE_JAR \
       -DpomFile=PATH_TO_FILE_POM
   ```

3. one Ontop developer must also create a GitHub token having `read:packages` as only permission

4. using such developer username and token, create file `.mvn/local-settings.xml`

   ```xml
   <settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
             xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
             xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0
                                 https://maven.apache.org/xsd/settings-1.0.0.xsd">
       <localRepository/>
       <interactiveMode/>
       <usePluginRegistry/>
       <offline/>
       <pluginGroups/>
       <servers>
           <server>
               <id>github</id>
               <username>USERNAME</username>
               <password>&#98;&#56;...XML-ENCODED READ-ONLY TOKEN...&#50;&#99;</password>
           </server>
       </servers>
       <mirrors/>
       <proxies/>
       <profiles/>	
       <activeProfiles/>
   </settings>
   ```

5. to use `local-setting.xml` by default in Maven, create file `.mvn/maven.config` with content:
   ```
   -s .mvn/local-settings.xml
   ```

6. in Ontop `pom.xml`, change ID and URL of custom repository to `github` and `https://maven.pkg.github.com/ontop/ontop` respectively; after that, any user (not just Ontop developers) could be able to run Maven on Ontop and have third-party JARs fetched from GitHub Packages.

ghxiao:
Hi @fracorco , thanks for the PR.

> 1. Decide on repository name `ontop-maven-repo`. Previously we discussed using `third-party-jars`, but I realized that this repository may be also used to deploy Ontop-related artifacts (e.g., Ontop snapshots not ready for Maven central, or the JARs of some other project under the ontop team that is not meant/ready to go on Maven central), so I switched to a more general name.

I would like to make this repo dedicated for dependencies. If we need to share Ontop snapshots for instance later, we can setup a separate repo for that or use the Github Packages services. Regarding the name, how about `ontop-3rd-party-maven-repo`? It is a bit long but reflects better the nature of this Github repository.

> 2. Read the [README.txt](https://github.com/ontop/ontop-maven-repo/blob/master/README.md) of such repository, focusing in particular on the disclaimers about license and third-party use (the instructions have been tested and work fine). Note that if we pick a different repository name the content of the README file should be adapted correspondingly.

I reviewed it and it looks good. We may need to adapt the repo configuration if we agree on the above suggestion of naming.

> 3. Please be aware that this feature branch also include a somehow unrelated change, i.e., the pinning of `tomcat-embed-core` to a specific version in the root `pom.xml`. That was needed for some reason in order to satisfy the `maven-enforcer-plugin`. If you prefer, we can implement this change as a separate bug fix.

This is OK.

fracorco:
Hi @ghxiao, thanks for the clarifications.

The solution `ontop-3rd-party-maven-repo` + create new GIT repository should we need another maven repo is fine.

Just out of completeness, there's another setup you may consider:
* create a single GIT repository to use for Maven repos
* for each Maven repository you need (e.g., `third-party-jars`, `ontop-snapshots`, etc) create a separate branch and upload the respective artifacts there
* in the `pom.xml`, include in the repository URL the branch name you want to use (currently there is `master`)

Note I'm not advocating for this setup, just it's technically feasible and consistent with some instructions found online. I did not consider it just for the sake of implementing something simple. Now branches can be used also in the solution proposed in this PR if needed (e.g., create `feature/x` branch of this GIT repository to match `feature/x` of Ontop, in case new third-party dependencies should be added). The main difference would be that instead of having multiple GIT repositories with a master branch each, one Git repo for each Maven repo, here we would have one single Git repo hosting all Maven repos in different branches. Actually I have no preferences, but I just wanted to point out this possibility.

fracorco:
Hi @ghxiao, following the agreed renaming of the Maven repository, its URL has been updated in this branch and Maven build has been successfully tested. Everything seems ready for merging the PR.

PR #453, opened 2021-07-28
Fix "cli" profile build error when ontop-rdf4j-* WAR generation is disabled

fracorco:
**tldr; dependencies of ontop-distribution now depend on active profile, to avoid failures when RDF4J wars are not built and are not available online**

This issue was likely present in version 4.1 too but I didn't notice it as it occurs only in very specific conditions.

Cause:
* maven is called with `cli` profile enabled (e.g., by `client/docker/build-image.sh`)
* the `cli` profile (as well as `protege` and `release` ones) skip the generation of the ontop-rdf4j-* WAR files
* still, ontop-distribution was unconditionally depending on those WAR files (for the case profile `webapps` was activated)
* if Maven could not resolve these WAR files locally (i.e., in the scope of current build) it tries to download them from remote repositories
  * that was working in version 4.1, so I didn't realize the issue
  * that did not work right after version was changed to 4.2-snapshot, as there were no JAR/WARs uploaded online for this new version yet
  * that started working again this evening while testing for version 4.2-snapshot, as apparently artifacts were uploaded to an online Maven repo

Fix:
* I changed ontop-distribution pom.xml so that dependencies are controlled by active profile, so Maven does not look anymore for these WAR files in case their generation is not enabled by current profile

Additional changes:
* I disabled by default the generation of WAR files, which was still done with profile -Prelease (although they were not zipped in a 
  release ZIP file under ontop-distribution/target). WAR files are now generated only if profile webapps is enabled.
* I added "set -e" at the beginning of build-image.sh, to prevent the script from continuing execution if Maven (or other build step) fails

Note: currently, profiles do not affect the Ontop modules included in the build, so all modules are always considered. What changes is only which Maven plugins are run on each module. In general, expensive plugins are disabled in profiles where they are not needed, while inexpensive plugins run always. This may be changed in case.

### DIFF
--- a/build/distribution/pom.xml
+++ b/build/distribution/pom.xml
@@ -18,46 +18,34 @@
         <assembly.attach>false</assembly.attach>
     </properties>
 
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>${project.groupId}</groupId>
+                <artifactId>ontop-protege</artifactId>
+                <version>${project.version}</version>
+                <exclusions>
+                    <exclusion>
+                        <artifactId>*</artifactId>
+                        <groupId>*</groupId>
+                    </exclusion>
+                </exclusions>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
     <dependencies>
         <dependency>
             <groupId>${project.groupId}</groupId>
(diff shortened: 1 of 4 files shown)


bcogrel:
Thanks a lot Francesco! Very clear explanation

Issue #478, opened 2021-11-06
Time function extension - duration


apano-on:
### Time function extension - duration arithmetic

Implement functions in Ontop which perform arithmetic between timestamps and dates with the goal of supporting expressions which retrieve the duration in user specified units between two dates or datetimes e.g.:
```
now() - ‘1970-01-01’^^xsd:date 
```
```
?end - ?start < 5 (days)
```

The output of these functions should be durations in user-specified units. Tentative prefix for these functions will be ofn with namespace to be defined. The functions to be implemented are:

- [x] ofn:weeksBetween
- [x] ofn:daysBetween
- [x] ofn:hoursBetween
- [x] ofn:minutesBetween
- [x] ofn:secondsBetween
- [x] ofn:millisBetween

Support both xsd:dateTime and xsd:date.

Also investigate adding more specific datatypes related to [duration](https://github.com/w3c/sparql-12/wiki/Inventory-of-existing-extensions-to-SPARQL-1.1#extended-mathematics-operations-on-datetimeduration-datatypes). A similar implementation to what will be added is available [here](https://graphdb.ontotext.com/free/devhub/time-functions.html#durations-expressed-in-certain-units).

apano-on:
Adopted the following rules which comply with the graphDB/Ontotext implementation:
- Datatype. xsd:long
- Namespace: <http://www.ontotext.com/sparql/functions/>
- Namespace prefix: ofn
- Truncation for date/datetime (where possible). GraphDB/Ontotext uses truncation, i.e. 25 days and 23 hours and 59 minutes difference --> 25 days difference. Some DB engines
automatically round to highest integer so this is not always possible.
- No Daylight Savings Time (DST) (where possible). GraphDB/Ontotext does not account for DST the extra (+1) or less (-1) hour.

Below is a list of DB engines which we aim to support and the respective functions/methodology used to get the desired time arithmetic:

- H2: [TIMESTAMPDIFF](http://www.h2database.com/html/functions.html#datediff)
- PostgreSQL: No built in function. Among several possibilities solution selected is to use [DATE_PART](https://www.postgresql.org/docs/8.1/functions-datetime.html) with some
math arithmetic to achieve the desired results as in https://www.sqlines.com/postgresql/how-to/datediff
- MySQL: [TIMESTAMPDIFF](https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_timestampdiff)
  - Other notes: MySQL always seems to handle DST correctly.
[...]

PR #483, opened 2021-11-30
Implementation of time duration arithmetic

apano-on:
This pull request provides a first implementation of the time arithmetic functions as defined in the [Ontotext implementation](https://graphdb.ontotext.com/free/devhub/time-functions.html#durations-expressed-in-certain-units).

Main features:
* Namespace: http://www.ontotext.com/sparql/functions/
* Namespace prefix: ofn
* Datatype. xsd:long
* List of functions (all support xsd:datatime arguments, and the first two also xsd:date):
  - ofn:weeksBetween(start, end)
  - ofn:daysBetween(start, end)
  - ofn:hoursBetween(start, end)
  - ofn:minutesBetween(start, end)
  - ofn:secondsBetween(start, end)
  - ofn:millisBetween(start, end)

For further details and a short discussion please refer to #478


apano-on:
As discussed we will try to ensure the results of the function were consistent across DB engines by among other things trying to see whether unix time or other solutions could help. The following changes were applied with that in mind:
* PostgreSQL - EXTRACT EPOCH used for unix time. This ensures all values are calculated more consistently.
* Microsoft SQL Server - No unix time function available (indeed DATEDIFF is used to get unix time). Therefore adjustments were performed to DATEDIFF to avoid the automatic rounding up. E.g. if we seek the difference in hours between timestamp2 and timestamp1, if the total minutes in timestamp2 is lower than that in timestamp1, we subtract 1 (to account for rounding). The same logic applies for other units (e.g. hour in this example) and their respective subunits (e.g. minute in this example).

Moreover, following an investigation of other changes (either to improve precision or function consistency) I can report the following:
* DB2 - The [EXTRACT](https://www.ibm.com/docs/en/ias?topic=functions-extract) function can be used to calculate unix time but it is usually off by 1 in the case of milliseconds due to automatic rounding rather than truncation. Therefore, since a non-unix time solution was used for milliseconds I kept the other built-in {TIME_UNIT}_BETWEEN functions defined. The results are consistent even without unix time. Timezone support for DB2 is also missing.
* Oracle - No built-in function to get unix time found. The use of EXTRACT is reasonably consistent (same results) albeit the resulting SQL translation is unfortunately very long.
* Denodo - No built-in function for unix time found. Results for timestamps without timezone are correct.
* Dremio - The function to retrieve unix time has as its lowest unit seconds. For seconds and larger units the built-in TIMESTAMPDIFF works correctly so no change was necessary. Support for milliseconds is missing in Dremio.

No changes needed for H2, MySQL.

Some issues with conversion of constant TEXT input to TIMESTAMP/DATE for PostgreSQL and Dremio were also addressed.

If further changes are necessary or some other solutions are possible I am available to look into them.

PR #499, opened 2022-01-10
Upgraded to RDF4J 3.7.4 and OWLAPI 5.19.1

avataar:
Upgraded to RDF4J 3.7.4 and OWLAPI 5.19.1
* Removed deprecated RDF4J usage
* Removed explicit use of obscure RDF4J utility classes that will be gone in RDF4J 4
* Upgraded/pinned some minor libraries versions


test/docker-tests/src/test/java/it/unibz/inf/ontop/docker/ParallelScenarioManifestTestUtils.java
```diff
@@ -51,6 +50,7 @@
 import java.util.jar.JarFile;
 
 public class ParallelScenarioManifestTestUtils {
+	private static final QueryLanguage SERQL_QUERY_LANGUAGE = QueryLanguage.valueOf("SERQL");
```
avataar, line 53: SERQL will be removed in RDF4J 4. Getting the constant indirectly makes the codebase easier to compile with RDF4J 4 (albeit the tests relying on SERQL will not run).

engine/reformulation/core/src/main/java/it/unibz/inf/ontop/answering/reformulation/input/impl/RDF4JConstructQueryImpl.java
```diff
@@ -73,6 +73,8 @@ private static TupleExpr getFirstNonSliceExpression(TupleExpr expr) {
     }
 
     private static UnaryTupleOperator getFirstProjection(TupleExpr expr) {
+        if (expr instanceof QueryRoot)
```
avataar, line 76: QueryRoot exists in RDF4J 3.x but isn't used. In RDF4J 4 parsed queries always start from QueryRoot so we need to unwrap them (needed in multiple places).

test/docker-tests/src/test/java/it/unibz/inf/ontop/docker/ParallelScenarioManifestTestUtils.java
```diff
@@ -63,7 +64,7 @@ public static TestSuite suite(ParallelScenarioTest.ParallelFactory factory) thro
 		if ("jar".equals(url.getProtocol())) {
 			// Extract manifest files to a temporary directory
 			try {
-				tmpDir = FileUtil.createTempDir("scenario-evaluation");
```
avataar, line 66: Method removed in RDF4J 4 so replaced with a more standard approach that doesn't depend on any dependency's version.

mapping/core/src/main/java/it/unibz/inf/ontop/spec/fact/impl/FactExtractorWithSaturatedTBox.java
```diff
@@ -297,6 +297,7 @@ else if (e instanceof DataSomeValuesFrom) {
                 .filter(m -> m instanceof ObjectSomeValuesFrom)
                 .map(m -> (ObjectSomeValuesFrom) m)
                 .filter(m -> !m.getProperty().isInverse()) //Drop the inverse of itself generated by Ontop
+                .filter(m -> !m.getProperty().toString().contains("urn:AUX.ROLE")) // Drop auxiliary object property
```
avataar, line 300: Test was failing with NPE (null object in a triple) presumably because the newer OWL API processes things in a different manner. I saw that urn:AUX.ROLE is filtered elsewhere in a similar manner.

ghxiao:
Thanks again for the PR.

The latest development of Ontop v5 already moved to RDF4J 4.1.3 and OWLAPI 5.1.20. This PR thus can be closed.

PR #556, opened 2022-09-30
Update sql-default.properties for OracleDriver

thomasjtaylor:
Problem: Incorrect case for oracle.jdbc.driver.OracleDriver-metadataProvider.
Ontop with `oracle.jdbc.driver.OracleDriver` incorrectly uses the default `DBMetadataProvider`, and therefore doesn't correctly extract Oracle metadata.

Fix incorrect case for oracle.jdbc.driver.OracleDriver-metadataProvider.
was: `oracle.jdbc.Driver.OracleDriver-metadataProvider`
now: `oracle.jdbc.driver.OracleDriver-metadataProvider`

Workaround 1: add correct property to User Properties:
`oracle.jdbc.driver.OracleDriver-metadataProvider=it.unibz.inf.ontop.dbschema.impl.OracleDBMetadataProvider`

Workaround 2: use `oracle.jdbc.OracleDriver`


thomasjtaylor:
closes #557

PR #566, opened 2022-10-23
SparqlQueryExecutor: support N-Triples and N-Quads formats

DylanVanAssche:
N-Triples and N-Quads are widely used RDF formats, support those besides Turtle, RDF/XML, RDF/JSON, etc.
This is useful to perform `CONSTRUCT` queries and get the results as N-Triples and N-Quads as well.

Would be great to have this in a release :)


bcogrel:
Thanks a lot @DylanVanAssche for the PR! And sorry for not having merged it earlier.

It will be included in the next major release. In the meantime, it is already available in the [4.3.0-SNAPSHOT Docker image](https://hub.docker.com/repository/docker/ontop/ontop-endpoint).

Best,
Benjamin

PR #576, opened 2022-12-09
Extend the VKG with INSERT SPARQL rules

bcogrel:
This PR introduces a major feature: using SPARQL INSERT rules to extend the VKG.

This work has been sponsored by ADP, a big thank to them!

The first idea for this feature came from [this thread](https://groups.google.com/g/ontop4obda/c/91HE1o50RKM) on the Ontop mailing list.

## Problem
A lot of domain logic can be encoded in a mapping (for instance in R2RML) between a data source and a knowledge graph. However, this logic would have to be encoded for each data source, which is tedious, error-prone and hard to maintain.

Virtual knowledge graph engines like Ontop allows to move some domain logic away from the mapping by putting it in a lightweight RDFS or OWL-2-QL ontology. That way this logic becomes data source-independent and can be applied uniformly across all the data. However, these ontology profiles are not expressive enough to cover many common needs. As a result, too much domain logic still has to be encoded manually in the mapping when using Ontop.

This PR aims at adding support for rules in Ontop so as to specify more domain logic in a data source-independent manner. These rules should be applied in particular contexts, ideally in the form of named graphs. These rules should preserve the ability to translate SPARQL queries into regular SQL queries and should not therefore contain any form of recursion.

## Proposed solution
We propose to add support for rules expressed in the form of [SPARQL 1.1 INSERT](https://www.w3.org/TR/sparql11-update/#insert) queries to Ontop, as follows:

 - The rules will be used for augmenting the internal mapping handled by Ontop with more definitions. These rules will be runned once, in order, over the internal mapping that has been saturated with the ontology.
 - The order between the rules will be determined by a dependency graph automatically generated from the rule definitions. Dependency rules will be executed before dependent rules. For instance, if a rule produces triples with property P1, a rule using a triple with the property P1 to produce a triple with the property P2 will have to be run after the first rule. In that case, we say that the second rule depends on the first one.

### Restrictions
 1. Cyclic dependencies are not supported, so as to guarantee the translatability of SPARQL queries into SQL queries. The dependency graph is therefore expected to be a directed acyclic graph (DAG).
 2. Classes and properties used in the SPARQL INSERT rule, both in the body and in the head, must be constant. They may not be variables or be assigned by SPARQL BIND expressions. This strongly simplifies the generation of a dependency graph and eliminates ambiguities about which properties or classes to consider at a given time.
 3. The rules contained in the ontology will be only runned before the SPARQL INSERT rules, not after. This reduces the implementation effort for this proposal.

These INSERT queries will not be inserted in the data sources. They will simply be virtually inserted in the virtual knowledge graph. We propose to use INSERT queries rather than CONSTRUCT queries because only the former are able to generate quadruples, i.e. putting triples in named graphs.

DELETE clauses are not supported as the rules only aim at adding data, not removing.

### 1 - Move the SPARQL query translator and the query unfolder in a new module

These classes are in the `ontop-reformulation-core` module, as they were only used for answering SPARQL queries at runtime. 
Now that we want to also parse SPARQL queries for augmenting the mapping, we need to move these classes in a new module, called `ontop-kg-query`. This new module will be inherited by both `ontop-reformulation-core` and `ontop-mapping-core`.

- [x] Create the `ontop-kg-query` module
- [x] Move the classes
- [x] Create new checked exceptions so as not to break the existing ones that were sub-classes of `OntopReformulationException`
- [x] Insert `OntopKGQuerySettings` and `OntopKGQueryConfiguration` in the existing hierarchies and update dependency injection.

### 2 - RuleExtractor and RuleExecutor

 A `RuleExtractor` converts SPARQL rules into `IQ`s, with one `IQ` per triple/quad pattern in the INSERT clause of the rule.
 
A `RuleExecutor` unfolds the rule `IQ`s and add them to the mapping.

### 3 - Normalize BGPs in SPARQL INSERT clauses

BGPs in the SPARQL INSERT clause can contain blank-nodes that have to relabelled with automatically generated templates.

These templates make use of an unique constraint if available, or otherwise of all the variables available.

### 4 - Ordering rules and detecting dependency cycles

Builds a dependency DAG between rules and use it for ordering the rules, so that dependencies are run before dependents.

Applies the restrictions agreed.

### 5 - Support WITH clauses in rules

Currently, Ontop does not support `WITH` clauses (nor `FROM` and `FROM NAMED` clauses in SELECT queries). 
However it supports the `GRAPH` block, so users can at the moment rewrite manually their queries with `WITH` clauses into equivalent ones without them.

See https://www.w3.org/TR/sparql11-update/#deleteInsert

This issue introduces support for WITH, USING and USING NAMED in INSERT operations, as well as FROM and FROM NAMED in SELECT queries (same mechanism).


bcogrel:
For a concrete example, see the class [SparqlRuleEmployeeTest](https://github.com/ontop/ontop/blob/43a6a68f0c46a119f3328d232653dbee5b06bc0e/binding/rdf4j/src/test/java/it/unibz/inf/ontop/rdf4j/repository/rules/SparqlRuleEmployeeTest.java) and [its SPARQL rule file](https://github.com/ontop/ontop/blob/43a6a68f0c46a119f3328d232653dbee5b06bc0e/binding/rdf4j/src/test/resources/employee/employee-rules.toml).

PR #599, opened 2023-02-06
Ontop support for trino

DRovara:
Created Docker Containers for trino:
- `trino_pg`: Postgres database containing the lightweight test database with some structural changes required to make trino connect.
- `trino`: Trino-Postgres connector that exposes routes to the three lightweight test databases.
Also created all property files and similar required for testing, as well as the Trino test classes themselves.

Implemented the trino ontop connector, including trino-specific:
- `DBTypeFactory`
- `DBFunctionSymbolFactory`
- `SelectFromWhereSerializer` (because `OFFSET` and `LIMIT` require specific order in trino)
- `DBMetadataProvider` (because default resulted in crash as it uses indexes which are not supported by trino)
- `QuotedIDFactory`

Also changed the `DefaultSelectFromWhereSerializer` class by changing the `viewCounter` attribute to protected because it was required in child class.


bcogrel:
Thanks a lot Damian! It will take part of 5.0.2.

Issue #550, opened 2022-09-16
Support requested for Athena / PrestoDB / Trino


bcogrel:
See https://github.com/ontop/ontop/issues/462#issuecomment-1224764143

bcogrel:
Support for Trino added: #599

bcogrel:
Support for PrestoDB added: https://github.com/ontop/ontop/pull/601

bcogrel:
Support for Athena added: #602

PR #602, opened 2023-02-10
Feature/athena

DRovara:
Ontop support for Athena.

As Athena is closely related to Presto and Trino, a large part of the Trino/Presto implementations could be reused for this feature. 

Only a new DBFunctionSymbolFactory was required, because Athena uses true/false as boolean literals for SQL queries, but uses 1/0 as boolean literals in results - the AthenaDBFunctionSymbolFactory handles this problem.

As of version 2.0.35.1000, Athena's JDBC does not support the retrieval of the default schema name. Therefore, ontop currently does not support default schemas for Athena. To run the lightweight test cases, I created new obda files with fully qualified names for all accesses tables. This may be changed in the future, if Athena's JDBC starts supporting default schemas.


bcogrel:
Great, thanks a lot Damian! 

Third PR of the week, closes https://github.com/ontop/ontop/issues/550 .

PR #605, opened 2023-02-20
Feature/redshift

DRovara:
Implemented support for Amazon Redshift.

This also includes the feature to have arbitrary Properties in JDBC properties file, as it was required to pass AWS access keys.


bcogrel:
Thanks Damian!

PR #608, opened 2023-02-21
Cleaning up Trino/Presto/Athena/Redshift/DuckDB Adapters

DRovara:
Changed some parts of the implementations to clean up the code, including:

- added license files for all dialects
- removed unnecessary "user" and "password" lines in duckdb properties files
- removed some types from the type factory that were included even though they are not supported.
- changed some `supports[DATATYPE]()` flags in type factories that previously had incorrect values.
- changed implementation of `createRegularDBConcatFunctionSymbol()` method for trino, duckdb and redshift, as they used the wrong implementation previously.


bcogrel:
Thanks Damian!

PR #612, opened 2023-03-01
Union Lens Feature

DRovara:
Added support for the `UnionLens` lens type.

3 additional fields:
`unionRelations`: A set of relations to compute the union of.
`makeDistinct`: Boolean value determining if result should be made distinct.
`provenanceColumn`: If provided, adds an additional column in which the source location of each data row is included.

Also added high-level tests:
- An example usage scenario, testing if provenance columns are included correctly and if all items are included in the union.

And low-level tests:
- Unions of two or more tables or lenses are performed correctly
- "Negative" test cases of union lenses in a format that is not supported (i.e. attributes have different types/names)
- Test cases on the inference of unique constraints (if all union tables have a unique constraint on an attribute and there exists a provenance column, then the full union has one too)
- Other added integrity constraints

### DIFF
--- a/core/optimization/src/test/java/it/unibz/inf/ontop/iq/optimizer/UniqueConstraintInferenceTest.java
+++ b/core/optimization/src/test/java/it/unibz/inf/ontop/iq/optimizer/UniqueConstraintInferenceTest.java
@@ -165,7 +165,7 @@ public class UniqueConstraintInferenceTest {
         IQTree tree = IQ_FACTORY.createNaryIQTree(
                 IQ_FACTORY.createUnionNode(ImmutableSet.of(X, A, B)),
                 ImmutableList.of(child1, child2));
-        assertEquals(ImmutableSet.of(ImmutableSet.of(X), ImmutableSet.of(X, A)), tree.inferUniqueConstraints());
+        assertEquals(ImmutableSet.of(ImmutableSet.of(X)), tree.inferUniqueConstraints());
     }
 
     /**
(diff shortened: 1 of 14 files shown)


core/optimization/src/test/java/it/unibz/inf/ontop/iq/optimizer/UniqueConstraintInferenceTest.java
```diff
@@ -165,7 +165,7 @@ public void testUnion2() {
         IQTree tree = IQ_FACTORY.createNaryIQTree(
                 IQ_FACTORY.createUnionNode(ImmutableSet.of(X, A, B)),
                 ImmutableList.of(child1, child2));
-        assertEquals(ImmutableSet.of(ImmutableSet.of(X)), tree.inferUniqueConstraints());
+        assertEquals(ImmutableSet.of(ImmutableSet.of(X), ImmutableSet.of(X, A)), tree.inferUniqueConstraints());
```
bcogrel, line 168: @DRovara , actually we should not add the second unique constraint, as it is already implied by the first one.
Can you please have a look?

The same remark holds for the following tests

db/rdb/src/main/java/it/unibz/inf/ontop/dbschema/impl/json/JsonUnionLens.java
```diff
@@ -0,0 +1,441 @@
...
+            throws MetadataExtractionException {
+
+        QuotedIDFactory quotedIDFactory = dbParameters.getQuotedIDFactory();
+        RelationID relationId = quotedIDFactory.createRelationID(name.toArray(new String[0]));
+
+        IQ iq = createIQ(relationId, dbParameters, parentCacheMetadataLookup);
+
+        int maxParentLevel = extractMaxParentLevel(iq, dbParameters.getCoreSingletons());
+
+        if (maxParentLevel > 0)
```
bcogrel, line 89: @DRovara , please remove this test and the warning (not relevant for union lenses)

DRovara: Done!

bcogrel:
Thanks Damian!

PR #613, opened 2023-03-02
Feature/maven docker refactoring

fracorco:
This PR  revises the Maven and Docker setup addressing issues related to the Ontop Protégé plugin, notably by enabling/disabling compiling of Ontop Protégé code while keeping all the Ontop Maven modules enabled. Alternative branch [feature/maven-docker-refactoring-toggle-modules](https://github.com/ontop/ontop/tree/feature/maven-docker-refactoring-toggle-modules) enables/disables modules instead, which has some impact on IDE behavior. Please reject this PR and open one on that branch if you prefer the latter (apart this different behavior, the changes and notes reported next applies to both branches).

Changes to Maven setup:
* reorganization of manually activated profiles to better fit Ontop v5 reorganization of Protégé modules (note: `-Dmaven.test.skip` has now to be manually supplied, as doing it by default had/has weird effects)
  - `plugin` - enables compiling/testing the protege plugin
  - `asset-cli` - enables building the ontop-cli github 'asset', replacing former `cli` profile
  - `asset-plugin` - enables building the protege plugin OSGI bundle
  - `asset-protege` - enables building the protege ZIP/tar.gz bundles, replacing former `protege` profile
  - `release` - all of the above, plus generation of javadoc jars, source jars, and extra metadata that are excluded by default
* `ontop-protege` module having again the root Ontop module as parent, this way inheriting the latter configuration and reducing duplications and possible conflicts
* `ontop-protege-plugin` by default not compiled, tested, packaged, installed, deployed unless profile `plugin` is activated, to avoid build failure when calling `mvn compile` or `mvn test` due to `ontop-protege-dependencies:shaded` JAR not yet produced
  - if `plugin` profile is activated and `mvn compile` or `mvn test` are called, an error with an informative message is generated using the Maven Enforcer plugin (instead of obscure compile errors)
* Protégé ZIP/tar.gz bundles downloaded either from a Maven repository or from their release URLs (new default as not yet available on Maven Central), depending on property 'protege.bnd.source' and various property 'protege.bnd.xxx' that configure the URL where to download from (documentation in `ontop-protege` `pom.xml`)
* fixed `readme.html` URL in `protege/plugin/update.properties` and `<update-url>` in the plugin OSGI manifest (apparently ignored by Protégé)
* removed obsolete and non-functional `build-plugin.sh` (if needed, can easily replace with a new script leveraging the Maven setup)
* consistent use of Logback only as logging backend (log4j2 was included by `ontop-mapping-owlapi`), with versions of different SLF4 dependencies harmonized to 1.7.36
* better control of enabled/disabled Maven plugins through a `${profile.phase-prefix}` mechanism (value is either empty or `disabled-`) that prepended to the plugin configured phase will either enable or disable it (no need to mess with plugin-dependent `<skip>` parameters, sometimes not available)
* updated all Maven plugins to most recent versions
  - update of Maven Resource plugin has highlighted a charset issue (UTF16 vs UTF8) of a filtered resource file, now manually converted to UTF8: `test/docker-tests/src/test/resources/dump/stockexchange-create-mssql.sql`
  - update of Maven Enforcer plugin has reduced build times of 10-15s
  - update of Antlr4 plugin has required updating the corresponding runtime dependency to the same latest version (only dependency updated, unit tests work)
* added Maven Enforcer rule to 'ban' inclusion of vulnerable log4j2 < v2.17 dependencies (build will fail if detected)
* removed profile `thirdparty` keeping plugins' configuration and operating only in `ontop-distribution`, in its place run
  - `mvn -Pthirdparty` => `mvn license:add-third-party`
* removed profile `analyze` keeping plugins' configuration, in its place run one or more (depending on needs) of
  - `mvn org.codehaus.mojo:versions-maven-plugin:display-plugin-updates`
  - `mvn org.codehaus.mojo:versions-maven-plugin:display-dependency-updates`
  - `mvn dependency:analyze-duplicate`
  - `mvn dependency:analyze-dep-mgt`
  - `mvn dependency:analyze`
* removed profile `format` keeping plugins' configuration, in its place run one or both of
  - `mvn com.github.ekryd.sortpom:sortpom-maven-plugin:sort`
  - `mvn net.revelc.code.formatter:formatter-maven-plugin:format`
* various minor cleanup and polishing of `pom.xml` files, plugin configurations (e.g., shade plugin to address some warnings) and assembly descriptors

Changes to Docker setup:
* `build-image.sh` and `Dockerfile` aligned to new Maven setup
* `build-image.sh` by default tags the generated image as `ontop/ontop:[version]` and `ontop/ontop:latest`, this behavior configurable with command-line option `-t` (see `build-image.sh --help`)
* `ontop` script kept unmodified and added to the PATH in generated Docker image

Remaining open issues:
* `ontop-protege-plugin` tests are still disabled due to incompatibility between OWLAPI 4.5.25 used in Protégé and the plugin, and OWLAPI 5.1.20 used in the rest of Ontop. No easy solution except maybe shading OWLAPI as well in the plugin and converting between shaded/unshaded OWLAPI objects.
* Maven 3.9 warning (does not happen with prior versions) about ```Parameter 'finalName' is read-only, must not be used in configuration``` when configuring the Assembly plugin with `finalName` to control the name of generated assembly file. Apparently this behavior was forbidden by Maven 2 but has always worked in Maven 3 due to a [bug](https://issues.apache.org/jira/browse/MNG-5001). Upcoming Maven 4 might (not decided yet) forbid this again thus breaking Ontop build. Suggested workaround of setting a unique `finalName` for the whole Maven module substantially impact the way we build and name Ontop CLI and Protégé asseblies. With some luck an updated version of the Assembly plugin will allow setting the output name without passing through `finalName`, otherwise we might have to change some naming convention in generated assemblies


fracorco:
wrongly opened against version5, will reopen

PR #615, opened 2023-03-03
BigQuery support for ontop

DRovara:
Added support for BigQuery.
BigQuery does not support integrity constraints, so tests for them were disabled.
Implemented new normalizer that pushes construct nodes into order by nodes if necessary.

Currently, the Simba BigQuery JDBC is not available in the maven central repository. Furthermore, their license does not allow us to publish it on the ontop maven repository. Therefore, tests for it have been commented out in the workflow, but may be run locally.

### DIFF
--- a/test/lightweight-tests/pom.xml
+++ b/test/lightweight-tests/pom.xml
@@ -547,6 +547,11 @@
             <version>1.12.251</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.google.cloud</groupId>
+            <artifactId>google-cloud-bigquery</artifactId>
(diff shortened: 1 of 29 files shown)


test/lightweight-tests/pom.xml
```diff
@@ -547,6 +547,12 @@
             <version>1.12.251</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
bcogrel, line 550: Please comment this dependency as it is unfortunately not public. It makes the CI fail.

bcogrel:
Thanks Damian!

PR #618, opened 2023-03-08
Added lightweight tests for Denodo and fixed some bugs in its implementation

DRovara:
- Added lightweight tests
- Did not add it to Github testing CI, because neither the Denodo Docker image, nor the Denodo JDBC are publicly accessible
- The directory lightweight-db-test-images/denodo instead contains database dumps exported from Denodo Express and a short guide on how to import them for testing.
- Added new normalizers because Denodo has the following special restrictions:
    - Order by clauses cannot contain complex expressions => instead use variable references
    - Limits cannot appear in sub-queries => remove them from sub-queries where valid
    - "IS NULL" cannot be called on disjunction/conjunction expressions => replace these expressions with "CASE" instead. 

All lightweight test cases that are not disabled in the corresponding classes passed during local testing.


bcogrel (approved):

bcogrel:
Thanks Damian!

PR #623, opened 2023-03-21
Replaced the UUID constant function call with RAND

DRovara:
We need to use some constant value cast to the target type as a workaround, so that Dremio does not optimize away the cast on NULL constants.
Previously, we used the UUID() function to get this constant value, as Dremio is less likely to optimize that compared to just using 0. 
However, UUID() is not supported by all Dremio versions. We now use RAND() instead to achieve the same goal.


bcogrel:
Was a problem for Dremio Cloud.

Thanks Damian!

PR #624, opened 2023-03-21
Added support for Dremio with the ArrowFlight JDBC Driver

DRovara:
It works out of the box by just assigning the DremioDBFunctionSymbolFactory, TypeFactory, MetadataProvider etc. to ArrowFlight.

In the future, if other types DBMS accessed through ArrowFlight are to be supported, this may need to be changed.


bcogrel:
Thanks!

PR #628, opened 2023-03-27
FlattenLens Postgres Arrays

DRovara:
Added support for Postgres array types `T[]` in FlattenLenses.

Also rewrote the lightweight test cases for nested data, so that they cover a wider range of corner cases and have fewer JSON-related examples.


bcogrel (approved):

bcogrel:
Thanks Damian!

PR #629, opened 2023-03-28
Support for FlattenLens on Dremio Lists.

DRovara:
Implemented FlattenLens for Dremio.

Currently unsupported features:
- Flatten with index variable is not supported by Dremio
- JSON Object access in basic lens expressions is currently not supported by ontop. Could be implemented at a different point in time.


bcogrel:
Thanks!

PR #630, opened 2023-03-28
XSD Cast Functions

apano-on:
This PR introduces support for cast functions from the [XPath Constructor Functions](https://www.w3.org/TR/sparql11-query/#FunctionMapping) section of the SPARQL 1.1 recommendation. 

Support is only extended for primitive types as per the matrix in the recommendation and date, resulting in the following new matrix (where always allowed (Y), never allowed (N) and dependent on the lexical value (M)):

image

The respective behavior of the functions strives to support the scenarios detailed in https://www.w3.org/TR/xpath-functions-31/#casting. However, due to the limitations of most SQL dialects, full compliance is not possible. Hence the following exceptions apply:
- `xsd:boolean`. Casting NaN to boolean should result in `"false"^^xsd:boolean`, however this check is not feasible for many DB dialects, hence there is mostly no special handling of these values with support limited to PostgreSQL, Oracle and Spark.
- `xsd:decimal`. Casting to DECIMAL results in values with scale 0 in several SQL dialects. As such the custom DECIMAL_STR with an arbitrary precision defined in the respective DBTypeFactory of a dialect is used for each cast.
- `xsd:dateTime`. The list of possible timestamp patterns is too large, and many dialects lack an effective way to check for these patterns without getting extremely verbose. Hence, it is not possible to often detect whether a value is a valid timestamp before the cast is performed. With the exception of SQL Server and Snowflake which have TRY_CAST functions, applying this function to other dialects comes with severe limitations.

### DIFF
--- a/binding/owlapi/src/test/java/it/unibz/inf/ontop/owlapi/BindTest.java
+++ b/binding/owlapi/src/test/java/it/unibz/inf/ontop/owlapi/BindTest.java
@@ -336,27 +336,23 @@ public class BindTest extends AbstractOWLAPITest {
     }
 
 
-    //CAST function not supported
-    //The builtin function http://www.w3.org/2001/XMLSchema#string is not supported yet!
-    @Test(expected = OntopReformulationException.class)
+    // NOTE: In H2 v2, we are forced to use DECIMAL(20, 6) because DECIMAL implies DECIMAL(*,0)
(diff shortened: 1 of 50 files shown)


bcogrel (approved):

bcogrel:
Thanks a lot Albulen for this great PR!

PR #631, opened 2023-03-29
Ontop property to ignore invalid mapping entries

DRovara:
Individual mapping entries with errors in them can now be ignored by setting the property `ontop.ignoreInvalidMappingEntries`

Also added a test case for this feature.


bcogrel (approved):

bcogrel:
Thanks Damian!

PR #632, opened 2023-03-29
Fixed an issue with dremio, where VALUES clauses with integer constants caused the query to fail

DRovara:
Achieved this, by changing the query serialization, so that integer constants in "Values" Nodes would be encased by CAST(<constant> AS INTEGER)


bcogrel:
Context: as of Dremio 24, constant integers are typed as `int64` in `VALUES` and as `int32` in `SELECT` clauses, which is inconsistent.

This PR overcomes this inconsistency.

bcogrel (approved):

bcogrel:
Thanks!

PR #635, opened 2023-04-03
FlattenLens implementation for Trino, Presto, and Athena

DRovara:
Implemented the flatten lens for the Trino family.

Notable points:
- the array type is called ARRAY(T) in Trino and Presto, but ARRAY<T> in Athena
- the JDBC correctly returns the type ARRAY(T) for columns in Trino and Presto, but only ARRAY for Athena, so additional casts need to be implemented to achieve the same results.
- Presto seems to be incompatible with Postgres arrays, so it was not possible to test the implementation for Presto. However, the Presto connector uses the same DBFunctionSymbolFactory and DBTypeFactory as Trino, and the "unnest" function in the documentation is described exactly the same as for Trino, so it is reasonable to assume it should work (though future testing would still be valuable)
- Changed the `BasicSingleTermTypeExtractor` to infer the flatten output type based on the flattened column type of the JoinLens, rather than the inferred input type, as it is more accurate this way.


bcogrel (approved):

bcogrel:
Thanks!

PR #636, opened 2023-04-03
Implemented Support for the Array/Object Access expression in Dremio

DRovara:
Expressions of the form array[key] can now be used in lenses for Dremio.
Currently, this expression is untyped. In the future, we may be able to create typed alternatives of it based on the type of the arrays used with it, but this is not possible for Dremio, as we do not know the exact types of Lists and Structs there.


bcogrel (approved):

bcogrel:
Thanks Damian!

PR #637, opened 2023-04-05
Implemented a common base structure for FlattenNode implementation

DRovara:
Many dialects use similar structures for the "flatten" procedure.
We can reduce duplicate code by implementing a common base structure for all dialects that use a flatten-like function in the SELECT parts of the query.


bcogrel (approved):

bcogrel:
Thanks Damian!

PR #638, opened 2023-04-11
Implemented FlattenLens for BigQuery

DRovara:
BigQuery supports the ARRAY type with type signature ARRAY<T>
BigQuery also supports the JSON type.

BigQuery does not support multi-dimensional arrays of the type ARRAY<ARRAY<T>>. However, we can have nested arrays of the form ARRAY<STRUCT<ARRAY<T>>> (this is the recommended approach from the BigQuery documentation).
The test cases have been implemented with this in mind.

The Flatten operation can be performed as a CROSS JOIN with the table function UNNEST [WITH OFFSET]. The `ordinality` column generated this way starts at 0 (whereas most other dialects start at 1).

Automated Github CI tests do not work for BigQuery, but all test cases were run locally to ensure nothing was broken.


bcogrel:
Thanks!

PR #639, opened 2023-04-11
Implemented FlattenLens for MariaDB, MySQL and MSSQL

DRovara:
MariaDB and MySQL work almost exactly the same (with very small implementation differences)

SQLServer uses the same syntax as MySQL, just function names are different, so it is included in this pull request.

These dialects only support the "flatten" function on JSON arrays (MySQL and MariaDB have a `JSON` data type, in SQLServer they are instead encoded in `VARCHAR` columns).

Due to what seems to be a bug in the implementation of MySQL, we need to encase the argument of the JSON_TABLE function (MySQL's and MariaDB's flatten) in a JSON_ARRAY(...) call and then deference this array again. In MariaDB, this workaround causes an error, which is why the dialects now have different SelectFromWhereSerializers.

For SQLServer, the position argument is only supported for the "serverless SQL pool in Synapse Analytics", which is why the position variable is disabled here (having a FlattenLens with a position column will throw an exception).


bcogrel:
Thanks Damian!

PR #640, opened 2023-04-11
Implemented FlattenLens for Oracle DB

DRovara:
Oracle has two categories of nested data types: Arrays/Structs/Maps and JSON objects (encoded as varchar2).

Arrays/Structs/Maps are quite complex as they seem to require the instantiation of custom user types and require several additional steps to be flattened.

For JSON arrays, the flatten procedure is more straightforward. Implemented the FlattenLens for JSON arrays only.

Can be implemented using the JSON_TABLE function (very similar to MySQL). Also supports position argument.

Unfortunately, there is no straightforward way to implement an `IS_ARRAY(...)` function for JSON values in Oracle. For now, we instead just check if the JSON string encoding starts and ends with '[' and ']' respectively. At a later point, the IS_ARRAY function may be implemented using Oracle's `JSON_DATAGUIDE(...)` function.

When arrays of strings are flattened, these strings will be encased in double quotes ("). A new normalizer was added to add a construction node over all flatten nodes for Oracle, which removes these quotes where necessary.


bcogrel (approved):

bcogrel:
Thanks one more time!

PR #641, opened 2023-04-13
Implemented FlattenLens for Redshift

DRovara:
Redshift uses the `SUPER` type for all nested data structures (arrays, maps, and structs). We cannot get more precise typing information, so we cannot infer the base type of an array (also, redshift arrays do not need to be all of the same type, so such a base type does not exist anyways).

Redshift supports the "ARRAY_FLATTEN" function, but this function flattens the array on all levels, so it is not compatible for our FlattenLens.

Instead, redshift also allows using array columns as `FROM` sources, so we perform the flatten operation as `SELECT <variables> FROM <subquery>, <array_column> AT <index_variable>`


bcogrel:
Thanks!

PR #642, opened 2023-04-14
Implemented FlattenLens for Snowflake

DRovara:
Snowflake uses the table function `FLATTEN` to unnest arrays.
Snowflake supports the datatype `ARRAY`, but it is not based on a single data type; arrays can contain values of any types.

It is not possible to use quotation marks in the alias of the FLATTEN results. Therefore, index and output variables have to be transformed to upper case and cannot contain special characters.


bcogrel:
Thanks Damian!

PR #643, opened 2023-04-17
Flatten Lens: Infer Unique and Foreign Key Constraints

DRovara:
Added functionality to infer integrity constraints for the Flatten Lens:
- UC: If a position argument `pos` is included in the Lens, and any of the base relation's UCs `c` is fully included, then `c` + `pos` is a unique constraint.
- FK: If any of the base relation's UCs is fully included, then we can use it as a foreign key to the base relation.


bcogrel:
Thanks!

PR #644, opened 2023-04-17
QuotedIDFactory refactoring introducing type annotation and lookup by type

fracorco:
This PR introduces a mechanism to link a `QuotedIDFactory` to a *factory type* string via [`QuotedIDFactory.@IDFactoryType`](https://github.com/ontop/ontop/blob/feature/quoted-id-factory-refactoring/core/model/src/main/java/it/unibz/inf/ontop/dbschema/QuotedIDFactory.java#L127) annotations, and to lookup a `QuotedIDFactory` instance via a [`QuotedIDFactory.Supplier`](https://github.com/ontop/ontop/blob/feature/quoted-id-factory-refactoring/core/model/src/main/java/it/unibz/inf/ontop/dbschema/QuotedIDFactory.java#L147) interface obtainable via injection.

`QuotedIDFactory.Supplier` injection leverages the a refactoring of [`OntopAbstractModule`](https://github.com/ontop/ontop/blob/feature/quoted-id-factory-refactoring/core/model/src/main/java/it/unibz/inf/ontop/injection/impl/OntopAbstractModule.java) that now allows the binding (overridable in settings file) of an implementation class to an arbitrary Guice *<class, qualifier>* key (instead of an *abstract class* key only), where the qualifier denotes here the factory type and allows having multiple coexisting `QuotedIDFactory` bindings for different type qualifiers. Specifically, `QuotedIDFactory` implementation classes are bound in [`OntopSQLCoreModule`](https://github.com/ontop/ontop/blob/feature/quoted-id-factory-refactoring/db/rdb/src/main/java/it/unibz/inf/ontop/injection/impl/OntopSQLCoreModule.java#L70) while an implementation of `QuotedIDFactory.Supplier` delegating to Guice `Injector` is bound in [`OntopModelModule`](https://github.com/ontop/ontop/blob/feature/quoted-id-factory-refactoring/core/model/src/main/java/it/unibz/inf/ontop/injection/impl/OntopModelModule.java#L128).

Note that the extension of `OntopAbstractModule` is general and can be used for similar use cases. Two unit tests are provided in [`OntopAbstractModuleTest`](https://github.com/ontop/ontop/blob/feature/quoted-id-factory-refactoring/core/model/src/test/java/it/unibz/inf/ontop/injection/impl/OntopAbstractModuleTest.java) and [`QuotedIDFactoryInjectionTest`](https://github.com/ontop/ontop/blob/feature/quoted-id-factory-refactoring/core/model/src/test/java/it/unibz/inf/ontop/injection/impl/QuotedIDFactoryInjectionTest.java), respectively for the handling of generic *<class, qualifier>* keys and its use for `QuotedIDFactory.Supplier`.

A `QuotedIDFactory.Supplier` is now injected into [`JsonMetadata`](https://github.com/ontop/ontop/blob/feature/quoted-id-factory-refactoring/db/rdb/src/main/java/it/unibz/inf/ontop/dbschema/impl/json/JsonMetadata.java#L93) when deserializing from JSON (via Jackson injection from [`JsonSerializedMetadataProvider`](https://github.com/ontop/ontop/blob/feature/quoted-id-factory-refactoring/db/rdb/src/main/java/it/unibz/inf/ontop/dbschema/impl/JsonSerializedMetadataProvider.java#L96)). This allows obtaining a `QuotedIDFactory` instance for the factory type occurring in the JSON by ultimately delegating to Guice (where knowledge about object instantiation resides), instead of hard-coding a *<type, implementation class>* `BiMap` and directly using `Class.newInstance()`.

Other changes in the PR are related to minor cleanup of other involved files (e.g., simplified binding expression in `OntopReformulationPostModule`, nullability handling in `QuotedIDImpl`) and in the use of a `@NonNullByDefault` annotation (taken from [io.github.solf:nullanno](https://github.com/solf/nullanno) just not to define it in Ontop code) to simplify and improve consistency - and thus improve IDE nullability analysis - of null-related annotations: when used on a class/package (as in classes modified in the PR), everything has to be non-null except where otherwise specified with `@Nullable`.

### DIFF
--- a/core/model/pom.xml
+++ b/core/model/pom.xml
@@ -62,6 +62,10 @@
             <artifactId>guice-assistedinject</artifactId>
         </dependency>
         <dependency>
+            <groupId>io.github.solf</groupId>
+            <artifactId>nullanno</artifactId>
+        </dependency>
+        <dependency>
             <groupId>com.google.code.findbugs</groupId>
             <artifactId>jsr305</artifactId>
         </dependency>
--- a/core/model/src/main/java/it/unibz/inf/ontop/dbschema/QuotedIDFactory.java
+++ b/core/model/src/main/java/it/unibz/inf/ontop/dbschema/QuotedIDFactory.java
@@ -9,9 +9,9 @@ package it.unibz.inf.ontop.dbschema;
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *      http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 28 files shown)


bcogrel (approved):

DRovara:
The test cases that are failing for athena never actually worked, as they are currently not supported. On version5, they are disabled.
Added the disabled tag to them here as well.

PR #646, opened 2023-04-20
Added STRING type to DuckDB

DRovara:
DuckDB also supports the type `STRING`.
This is useful for the Lens tutorial, so I would like to include it.

(I also removed some unused imports from the NestedDataDuckDBTest class)


bcogrel:
Thanks!

PR #647, opened 2023-04-21
Replaced 'POSITION' function with 'INSTR' for SparkSQL

DRovara:
Databricks and SparkSQL support both the function `INSTR` and `POSITION` to find the position of a substring in a different string. However, Palantir Foundry only supports `INSTR` (https://www.palantir.com/docs/foundry/transforms-sql/spark-reference/#string-functions)

Therefore, the instances of `POSITION` were replaced with `INSTR` for Spark.


bcogrel:
Thanks Damian!

PR #648, opened 2023-04-21
Fixed incorrect implementation of "Ignore Invalid Mapping Entries" functionality

DRovara:
Now, there are two try-catch blocks:
  - One large try-catch block for the out for loop to make sure all issues found in the processing of the mapping in general are covered (same as before)
  - One try-catch block inside the inner loop that catches issues on individual target atoms and only ignores those failing atoms (new)


bcogrel:
Thanks!

PR #650, opened 2023-04-28
Fixes #649

### LINKED ISSUE #649: Docker image can't read command line arguments from files
### Description

The Docker `entrypoint.sh` script uses the bash specific `$(< "${ONTOP_DB_PASSWORD_FILE}")` notation but the shebang path was changed to `/bin/sh` [here](https://github.com/ontop/ontop/commit/149a0786e69e51ca2d7f726f2247ed5fc700a5ad#diff-e70a7161b27a48cc780b54124b68c0e40d946ca406fca49418307fda8b12a594L1).

### Steps to Reproduce

1. Try to run the Docker container with the `ONTOP_DB_PASSWORD_FILE` environment variable set.

**Expected behavior:** Password is read from the file specified by `ONTOP_DB_PASSWORD_FILE`.

**Actual behavior:** No password is read in and the container fails to start

**Reproduces how often:** Whenever the `entrypoint.sh` script tries to read an argument from a file specified by a `_FILE` environment variable.

### Versions

Docker image version 5.0.0+

### Additional Information

Changing the shebang back to `#!/bin/bash` seems to fix the issue.

fracorco:
Restores capability of reading DB url, user, password from files, broken due to use of `bash` syntax with `sh` entrypoint interpreter. Fixed by using `cat` and remaining compatible with `sh`, so to allow possible use of Alpine base Docker image that does not include `bash`.

### DIFF
--- a/client/docker/entrypoint.sh
+++ b/client/docker/entrypoint.sh
@@ -76,7 +76,7 @@ if [ "${ONTOP_DB_USER_FILE+x}" ]; then
   if [ "${ONTOP_DB_USER+x}" ]; then
     echo "ERROR: environment variables ONTOP_DB_USER and ONTOP_DB_USER_FILE are conflicting. Please choose one of the two." && exit 1
   fi
-  set -- "$@" "--db-user=$(< "${ONTOP_DB_USER_FILE}")"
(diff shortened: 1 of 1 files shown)


bcogrel:
Thanks Francesco!

PR #651, opened 2023-05-09
Fixes issue in RDF4JTupleExprTranslator that made UNIONS fail

DRovara:
SPARQL queries with UNION operations were failing under certain conditions.
This problem is rooted in the `translate(Union union)` method in `RDF4JTupleExprTranslator`. 

The method creates a projection over each of the two children of the UNION operation so that missing columns are added to the respective child. 
Afterwards, it checks if the sub-tree of each child contains a non-projected variable with the same name. If so, it renames all occurrences of that variable in that sub-tree. However, because the operations occur in this order, it will also rename the newly added column that was required for the UNION, and the query will fail.

Fixed the issue by changing the order of these steps: First we rename the variable occurrences inside the sub-tree and then we add the projection that adds the missing variables.


bcogrel:
Thanks Damian!

PR #652, opened 2023-05-09
Possible fix for "CONSTRUCT queries with unions and different variables return empty results" issue

DRovara:
This is a bugfix for the issue on the ontop-issues GitLab repository "CONSTRUCT queries with unions and different variables return empty results".

The bug was caused by the following procedure. Imagine a UNION of the type
```SPARQL
{ }
UNION
{ ?s ?p ?o . }
```
The way UNIONs work in SQL, all operands of a UNION need the same columns. Therefore, null-values are added to the first set, and then the UNION is performed.
Because of this, the first row of the SQL result set will be a row of NULLs.

Now, the results will be stored as a `DefaultSimpleGraphResultSet` by ontop. When calling `next()` on this result set, it will access the first row of the SQL result first. Because of this, all values will be null in the first try, so this row will not be added to the `statementBuffer`. However, this now means that the `statementBuffer` will remain empty, and because of that, `hasNext()` will return `false`, and it will appear as if there is no result at all. 

I fixed this by adding a `do ... while` in the `next()` call of the `DefaultSimpleGraphResultSet`. The procedure will repeat if the `statementBuffer` has been left empty even though the SQL result set is not empty.

Unfortunately, I am not fully sure if doing this would cause any unwanted consequences, as I do not have a full overview of the exact tasks the `DefaultSimpleGraphResultSet` is supposed to perform

### DIFF
--- a/binding/rdf4j/src/test/java/it/unibz/inf/ontop/rdf4j/repository/AbstractRDF4JTest.java
+++ b/binding/rdf4j/src/test/java/it/unibz/inf/ontop/rdf4j/repository/AbstractRDF4JTest.java
@@ -189,6 +189,26 @@ public class AbstractRDF4JTest {
         assertEquals(expectedGraph, statementBuilder.build());
     }
 
+    protected int runGraphQueryAndCount(String queryString) {
+        return runGraphQueryAndCount(queryString, new MapBindingSet());
+    }
+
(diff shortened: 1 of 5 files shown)


bcogrel (approved):

bcogrel:
Good catch! Thanks Damian!

PR #656, opened 2023-05-12
Fixed issue that made lenses lose UCs under certain conditions

DRovara:
With the current implementation of the `LensMetadataProviderImpl` class, there are cases when lenses could "lose" their integrity constraints if they are in a multi-level hierarchy of lenses. This can happen under the following circumstances:

Imagine two basic lenses, `L1` and `L2`, where `L2` uses `L1` as a base relation, and `L1` has a table with one attribute `id` as its base. `id` is UNIQUE.

Now, ontop reaches the point where the integrity constraints are computed (`LensMetadataProviderImpl.insertIntegrityConstraints(...)`). It iterates over the list of lenses (*loop 1) and computes the constraints for each of them.
If it starts with `L2`, then the method will notice that it uses `L1` as its base relation. It will take the LensImpl object for `L1` from `L2`'s base relations and call `insertIntegrityConstraints(...)` on it first. Once this step is complete, the RelationID of `L1` is added to the `alreadyProcessedViews` list.
Next, *loop 1 continues and `insertIntegrityConstraints(...)` is called on the Lens `L1`. As the RelationID of `L1` is already in `alreadyProcessedViews`, it will not compute L1's integrity constraints again. However, this LensImpl object of `L1`, and the one taken from L2's base relation were _different instances_, because the `getRelation(...)` method in `LensMetadataProviderImpl` constructs a new `LensImpl` object each time it is called. Because of that, this `L1` instance will keep its integrity constraints empty.
If *loop 1 were to start with `L1` instead, the same thing would happen in the different direction: L2's base relation `L1` instance will not receive any UCs, and so `L2` cannot infer them.

I fixed this by adding a cache to `LensMetadataProviderImpl` so that `getRelation(...)` will not always construct new instances, and instead reuse existing instanses of `LensImpls` if `getRelation(...)` is called multiple times for the same RelationID.

### DIFF
--- a/db/rdb/src/main/java/it/unibz/inf/ontop/dbschema/impl/LensMetadataProviderImpl.java
+++ b/db/rdb/src/main/java/it/unibz/inf/ontop/dbschema/impl/LensMetadataProviderImpl.java
@@ -17,10 +17,7 @@ import it.unibz.inf.ontop.utils.ImmutableCollectors;
 import javax.annotation.Nullable;
 import java.io.IOException;
 import java.io.Reader;
-import java.util.Comparator;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
 import java.util.stream.Stream;
 
 public class LensMetadataProviderImpl implements LensMetadataProvider {
@@ -35,6 +32,8 @@ public class LensMetadataProviderImpl implements LensMetadataProvider {
 
     // "Processed": constraints already inserted
     private final Set<RelationID> alreadyProcessedViews = new HashSet<>();
+
+    private final Map<RelationID, NamedRelationDefinition> cachedLenses = new HashMap<>();
 
     // Lazy (built lately)
     @Nullable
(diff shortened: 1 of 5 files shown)


db/rdb/src/main/java/it/unibz/inf/ontop/dbschema/impl/LensMetadataProviderImpl.java
```diff
@@ -106,8 +105,12 @@ public ImmutableList<RelationID> getRelationIDs() throws MetadataExtractionExcep
     @Override
     public NamedRelationDefinition getRelation(RelationID id) throws MetadataExtractionException {
         JsonLens jsonLens = jsonMap.get(id);
-        if (jsonLens != null)
-            return jsonLens.createViewDefinition(getDBParameters(), dependencyCacheMetadataLookup.getCachingMetadataLookupFor(id));
+        if (jsonLens != null) {
+            if (!cachedLenses.containsKey(id))
```
kontchakov, line 109: computeIfAbsent?

bcogrel:
Thanks Damian!

Issue #593, opened 2023-01-31
Rdb2RdfTests D0014 Postgresql unique constraint fails for ontop-bnode creation (INT treated as TEXT)


thomasjtaylor:
# Description
Similar to #564 (also for Postgreql INT treated as TEXT without explicit cast)

Rdb2RdfTest D0014 creates three tables (`DEPT`, `EMP`, and `LIKES`).

`DEPT.deptno` is defined as `INTEGER UNIQUE` (as opposed to `INTEGER PRIMARY KEY`) which causes Ontop to generate b-nodes for each `DEPT` row/entity in the database.

- [DEPT Table Only](#dept-table-only) SUCCESS - appears to post-process the query results to generate b-nodes; does not use `REPLACE()`.
- [EMP Table Only](#emp-table-only) SUCCESS - no b-nodes, does not use `REPLACE()`.
- [DEPT+EMP Tables](#dept-emp-tables) FAILURE - appears to try to include the `ontop-bnode-1` statement in the query, uses `SELECT ('ontop-bnode-1' || ... REPLACE(v17."deptno", '%', '%25')` - since `deptno` is INTEGER, Postgres reports an error: `function replace(integer, unknown, unknown) does not exist` 
  - If I change `DEPT.deptno INTEGER UNIQUE` to `DEPT.deptno INTEGER FRIMARY KEY`, no b-nodes are generated and the query works.
    - [DEPT + EMP Tables with PK](#dept-emp-tables-with-pk) - SUCCESS - no b-nodes (since both EMP and DEPT have primary keys)
  - If I remove the `EMP.job VARCHAR(30)` column from the schema, the query works. _**Strange**, since b-nodes are for the DEPT table._
    - [DEPT + EMP Tables no Job](#dept-emp-tables-no-job) - SUCCESS - b-nodes are generated but doesn't use `REPLACE()`, instead query keeps INTs `(SELECT v3."deptno" AS "deptno1m10", CAST(NULL AS INTEGER) AS "deptno1m17"...`

### Problem Summary:
It appears that at some level of complexity, Ontop switches from implicitly adding b-nodes to explicitly including the 'ontop-bnode-N' statement in the query. At that point, the INTEGER fields shouldn't be treated as TEXT (e.g., remove all the REPLACE functions), or INTEGER fields should be cast to TEXT (although this would be less ideal).

### Test Code:
[...]

PR #657, opened 2023-05-12
Added missing cast functions to if-branch without primary keys for RDB2RDF

DRovara:
Fixes a problem in RDB2RDF where an integer variable used in an IRI of a bnode would fail, as it didn't cast the integer to a string first, making the variable incompatible with the REPLACE function used for IRI encoding. 

This error only occurred for tables that don't have a primary key, as the code branches depending on if there is a primary key or not. In the "else" case, it seems like the addition of casting functions was forgotten.

This should fix [Issue 593](https://github.com/ontop/ontop/issues/593)


bcogrel:
Well spotted!

PR #660, opened 2023-05-19
Implemented SAMPLE DB function symbol to improve performance

DRovara:
Added SAMPLE aggreagate function symbol for dialects that support it.
This is faster than relying on MIN (the current implementation)

Sample or similar functions are supported by:
Snowflake, Oracle, BigQuery, Redshift, Trino, Presto, and Athena, SparkSQL, DuckDB and Denodo

And not supported by:
DB2, Dremio, MySQL, MariaDB, SQLServer, and Postgres


bcogrel (approved):

bcogrel:
Great! Thanks a lot Damian!

PR #661, opened 2023-05-22
Added index type of FlattenNode as LargeInteger

DRovara:
Currently, the `getIndexVariableType()` method returns `Optional.empty()`.
However, this can cause problems: When used as part of an IRI template, the variable has to be cast to `STRING`. Since we don't know what the base-type is, it will use a `DefaultSimpleDBCastFunctionSymbol`. For this function symbol,
`isAlwaysInjectiveInTheAbsenceOfNonInjectiveFunctionalTerms()` returns false if the input type is unknown.

This way, the `analyzeInjectivity(...)` method in `FunctionSymbolImpl` will not return a decomposition; therefore, the `ConstructionNode` will not know that the Unique Constraint that may have been holding before the IRI construction will still hold, which prevents the query from being optimised.

Adding an expected type for the index variable prevents all of this, and the unique constraint will be carried over.

I chose `DBLargeInteger` as the type, as this is the most reasonable one. It works with all currently supported databases, and there is no reason to assume an index of the `FLATTEN` function for any different database may be of a non-integer type.

### DIFF
--- a/core/model/src/main/java/it/unibz/inf/ontop/iq/node/impl/FlattenNodeImpl.java
+++ b/core/model/src/main/java/it/unibz/inf/ontop/iq/node/impl/FlattenNodeImpl.java
@@ -98,7 +98,7 @@ public class FlattenNodeImpl extends CompositeQueryNodeImpl implements FlattenNo
 
     @Override
     public Optional<TermType> getIndexVariableType() {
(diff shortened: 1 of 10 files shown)


bcogrel:
Good catch! Thanks Damian!

PR #666, opened 2023-06-09
Added support for additional input file for facts

DRovara:
Currently, additional facts have to be passed to Ontop through the ontology file - this is not a clean solution.

Added new parameters to all Ontop CLI commands to also provide a "facts" file in any RDF format.
The following parameters have been added 

-a <file>; --facts <file>: A file containing a collection of facts.
--fact-format <format>: The format to be used to parse the file. If not provided, we try to infer the format from the file extension. Supports the same formats as the `ontop materialize` command.
--facts-base-iri <baseIRI>: The base IRI used for the provided fact file. If none is given, a random IRI is generated instead.

Also added test cases for fact files in RDFXML format, TURTLE format, and in N-QUADS format with additional context graphs.

### DIFF
--- a/binding/owlapi/src/test/java/it/unibz/inf/ontop/owlapi/AbstractOWLAPITest.java
+++ b/binding/owlapi/src/test/java/it/unibz/inf/ontop/owlapi/AbstractOWLAPITest.java
@@ -52,8 +52,14 @@ public class AbstractOWLAPITest {
         initOBDA(createDbFile, obdaFile, ontologyFile, propertiesFile, null);
     }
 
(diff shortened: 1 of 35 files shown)


bcogrel:
Thanks, it looks pretty good!

May I ask to:
  1. Add [environment variables for the Docker image](https://github.com/ontop/ontop/blob/version5/client/docker/entrypoint.sh) and update [its README](https://github.com/ontop/ontop/blob/version5/client/docker/README.md)
  2. Move the new unit tests from `ontop-owlapi` (legacy, we don't want to encourage users to create new tests there) to `ontop-rdf4j`?

bcogrel:
Thanks!

PR #668, opened 2023-06-14
Propagate Unique Constraints of lenses upwards

DRovara:
Basic implementation of the feature to push unique constraints of lenses upwards, implemented only for basic lenses for now.

If a basic lens has a UC added, it will try to add the same UC on the parent as well, if this is possible. This way, such UCs can be propagated all the way up to the base relation.


bcogrel:
Thanks!

PR #676, opened 2023-06-16
Fixed issue that made SQLServer fail queries that involved a DISTINCT and a LIMIT

DRovara:
Limit only works in the presence of ORDER BY, so `ORDER BY (SELECT NULL)` as  added to circumvent this. However, this is not valid if the query uses `DISTINCT`, as it will fail with `com.microsoft.sqlserver.jdbc.SQLServerException: ORDER BY items must appear in the select list if SELECT DISTINCT is specified.`

Changed the SelectFromWhereSerializer of SQLServer to use `ORDER BY 1` instead, which is supported even with DISTINCT.


bcogrel:
Thanks Damian!

PR #679, opened 2023-06-20
Fixed ORDER BY 1 issue for SQLServer

DRovara:
Previously, we changed the `ORDER BY (SELECT NULL)` queries to `ORDER BY 1`, as the original approach would fail in the presence of `DISTINCT`.

However, we figured out that this solution will not always work. In particular, if column number 1 is of type XML, the query fails.

To reduce the likelihood of the error coming up, we added an additional condition: `ORDER BY 1` is only used if a `DISTINCT` appears. This way, queries on tables starting with XML columns will not always fail, but only if they also require DISTINCT, which should largely reduce the likelihood of this issue appearing.

In a future implementation we may fully fix this problem by generating a new column that will be of a simple data type and performing the `ORDER BY` on that column.


bcogrel (approved):

bcogrel:
Thanks Damian!

PR #681, opened 2023-06-22
Functional Dependency Inference

DRovara:
Implemented a basic structure for the inference of functional dependencies on the IQTree.

The IQTree now provides the method `inferFunctionalDependencies()` to get a list of possible functional dependencies. It is not guaranteed to find all of them, but all dependencies found are guaranteed to be correct.

Some improvements can still be made, in particular in the context of the LeftJoin node, which currently only infers FDs from the left child.

At this point, the `inferFunctionalDependencies()` method is not called from anywhere. This may change in the future when we require its functionality.


bcogrel (approved):

bcogrel:
Great feature, thanks Damian!

PR #682, opened 2023-06-22
Implemented normalizer for SQLServer that adds an order by clause inside slices

DRovara:
SQLServer does not support LIMIT and OFFSET on rows that have not first been ordered.

Added a new normalizer that inserts ORDER BY nodes inside slices if they do not already have one. The ORDER BY node sorts on a newly created variable to guarantee that it can always be sorted successfully.

This means we no longer require the old "ORDER BY (SELECT NULL)" used in the `SQLServerSelectFromWhereSerializer`, as it caused issues.


bcogrel:
Thanks Damian!

PR #685, opened 2023-06-26
Push terms used in substitution below DISTINCT node, if they do not support distinct.

DRovara:
Implemented a flag for data types to indicate that they should not be used with `DISTINCT`.

Should a case appear where a function uses such a data type, the construction will be split into a part that is possible above DISTINCT and a part that must be below it, and Ontop will attempt to create a new construction node below the DISTINCT that contains the function call, so that the limitation may be circumvented.

Also supports cases with more complex function calls, (`F(G(X))`, `F(G(X), H(X))`, etc.).

The following conditions must hold for this transformation to be applied:
    - A functional dependency from another projected variable to the variables used in the pushed-down expression must exist.
    - The functions that are pushed down must all be deterministic.

### DIFF
--- a/core/optimization/src/main/java/it/unibz/inf/ontop/iq/optimizer/impl/PreventDistinctOptimizerImpl.java
+++ b/core/optimization/src/main/java/it/unibz/inf/ontop/iq/optimizer/impl/PreventDistinctOptimizerImpl.java
@@ -1,0 +1,218 @@
+package it.unibz.inf.ontop.iq.optimizer.impl;
+
+import com.google.common.collect.*;
+import it.unibz.inf.ontop.exception.MinorOntopInternalBugException;
+import it.unibz.inf.ontop.injection.IntermediateQueryFactory;
+import it.unibz.inf.ontop.injection.OptimizationSingletons;
+import it.unibz.inf.ontop.iq.IQ;
+import it.unibz.inf.ontop.iq.IQTree;
+import it.unibz.inf.ontop.iq.UnaryIQTree;
+import it.unibz.inf.ontop.iq.node.*;
+import it.unibz.inf.ontop.iq.optimizer.PreventDistinctOptimizer;
+import it.unibz.inf.ontop.iq.transform.IQTreeTransformer;
+import it.unibz.inf.ontop.iq.transform.impl.DefaultRecursiveIQTreeVisitingTransformer;
+import it.unibz.inf.ontop.model.term.*;
+import it.unibz.inf.ontop.model.type.DBTermType;
+import it.unibz.inf.ontop.model.type.TermTypeInference;
+import it.unibz.inf.ontop.substitution.SubstitutionFactory;
+import it.unibz.inf.ontop.utils.ImmutableCollectors;
+import it.unibz.inf.ontop.utils.VariableGenerator;
+
+import javax.inject.Inject;
+import java.util.Optional;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+public class PreventDistinctOptimizerImpl implements PreventDistinctOptimizer {
+
+    private final IntermediateQueryFactory iqFactory;
+    private final OptimizationSingletons optimizationSingletons;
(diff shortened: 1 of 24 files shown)


core/optimization/src/main/java/it/unibz/inf/ontop/iq/optimizer/impl/PreventDistinctOptimizerImpl.java
```diff
@@ -0,0 +1,225 @@
...
+            }
+            return super.transformConstruction(tree, rootNode, child);
+        }
+
+        private IQTree performTransformation(ConstructionNode constructionNode, IQTree originalTree, Decomposition decomposition) {
+            var subtree = decomposition.descendantTree;
+
+
+            //Find terms that need to be pushed below the distinct.
+            ImmutableMap.Builder<Variable, ImmutableTerm> substitutionBuilder = ImmutableMap.builder();
```
bcogrel, line 76: Please find a better name for it. It is not building a substitution object but a map. From its name, it is a bit hard to understand what this map is about.

core/optimization/src/main/java/it/unibz/inf/ontop/iq/optimizer/impl/PreventDistinctOptimizerImpl.java
```diff
@@ -0,0 +1,225 @@
...
+            if(substitutionBuilder.build().isEmpty())
+                return originalTree;
+            var substitution = substitutionBuilder.build();
+            if(!substitution.values().stream()
+                    .allMatch(this::isDeterministic))
+                throw new IllegalArgumentException("Unable to push non-supported terms below DISTINCT, because they include a non-deterministic function.");
+
+
+            //Check if all variables used by these terms have a functional dependency pointing to them.
+            var pushedVariables = substitution.entrySet().stream()
```
bcogrel, line 89: Let's perhaps use a substitution instead a map to avoid this boilerplate

core/optimization/src/main/java/it/unibz/inf/ontop/iq/optimizer/impl/PreventDistinctOptimizerImpl.java
```diff
@@ -0,0 +1,225 @@
...
+
+            //Construct the updated IQTree.
+            var topConstruction = iqFactory.createConstructionNode(
+                    constructionNode.getVariables(),
+                    updatedTerms.stream()
+                            .filter(entry -> !(entry.getValue() instanceof Variable))
+                            .collect(substitutionFactory.toSubstitution())
+            );
+            var bottomSubstitution = substitution.entrySet().stream()
+                    .map(entry -> Maps.immutableEntry(
```
bcogrel, line 107: Move this logic to the collector (no need to rebuild a Map.Entry). Please split in several lines for better readability

core/optimization/src/main/java/it/unibz/inf/ontop/iq/optimizer/impl/PreventDistinctOptimizerImpl.java
```diff
@@ -0,0 +1,225 @@
...
+                    .anyMatch(v -> functionalDependencies.getDeterminantsOf(v).stream()
+                            .noneMatch(determinants -> Sets.intersection(determinants, pushedVariables).isEmpty())))
+                throw new IllegalArgumentException(String.format("No Functional Dependency could be used to push the variables %s down.", pushedVariables));
+
+
+            //Construct the updated IQTree.
+            var topConstruction = iqFactory.createConstructionNode(
+                    constructionNode.getVariables(),
+                    updatedTerms.stream()
+                            .filter(entry -> !(entry.getValue() instanceof Variable))
```
bcogrel, line 103: What if there were already variable-to-variable entries in the original substitution?

bcogrel (approved):

bcogrel:
Great feature, thanks Damian!

PR #693, opened 2023-06-29
Added support for DynamoDB

DRovara:
Ontop can now be used together with DynamoDB using CData's DynamoDB JDBC:
https://www.cdata.com/drivers/dynamodb/jdbc/

Nested data structures and the flatten lens are currently not supported.

DynamoDB cannot be added to the testing pipeline, but it was tested locally.


bcogrel:
Thanks Damian!

PR #702, opened 2023-07-10
Added partial RegEx support to SQLServer

DRovara:
SQLServer does not directly support regular expressions and the `REGEXP_MATCH` function.

Implemented a new FunctionSymbol that can be used to translate REGEXP_MATCH calls into `(<expression> LIKE <pattern>)` calls, where `pattern` is derived from the regular expression pattern provided.

As the `LIKE` operator is quite primitive compared to RegEx, only a limited number of patterns are supported this way.

Notes:
- The arity-3 version of this function is also implemented, but only for the flag "i".
- `REGEXP_REPLACE` cannot be implemented in a similar fashion, as we would need to find and replace _any number_ of pattern occurrences to implement it, which cannot be done easily.


bcogrel:
Great, thanks a lot Damian!

PR #706, opened 2023-07-12
Typing for math operators used in the mapping / lenses

DRovara:
Implemented feature to support type inference for mathematical binary expressions used in mapping and lens files.
Implemented transformer that transforms un-typed math binary function symbols into types ones, if their types can be inferred during execution.


bcogrel (approved):

bcogrel:
Fantastic!

PR #709, opened 2023-07-18
Ignoring invalid mapping entries will now also ignore invalid lenses

DRovara:
When a lens with invalid properties (e.g. accessing invalid table, reading invalid column) is encountered and the ignoreInvalidMappingEntries property is set, such invalid lenses will now also be ignored.

The LensMetadataProvider will instead return an empty relation definition in those cases.


bcogrel:
Thanks Damian!

PR #713, opened 2023-07-20
Composite UC inference now only considers variables that are related to the UC

DRovara:
Previously, the time requirement of the inference procedure was exponentially dependent on the number of projected variables.

The new implementation makes sure only variables are considered that are related to the UC (i.e. were already previously part of the UC or are determined by a variable part of the UC). 
This way, it is only exponential (roughly) in the size of the UCs, which should substantially increase the performance in certain cases.


bcogrel:
Thanks Damian for the prompt fix!

PR #721, opened 2023-07-28
Fixed Old Version SQLServer LimitOffset normalization

DRovara:
Previously, the Limit and Offset normalizer for old versions of SQLServer did not apply the transformation to subqueries in case they contain another slice node.

Because of that, complex queries failed on old versions of SQLServer.

This pull request fixes this issue.


bcogrel:
Thanks Damian!

PR #735, opened 2023-08-10
Implemented basic statistical Math Aggregate functions

DRovara:
The following aggregate functions can now be used in SPARQL queries:
- `<http://jena.apache.org/ARQ/function/aggregate#stdev>`
- `<http://jena.apache.org/ARQ/function/aggregate#stdev_pop>`
- `<http://jena.apache.org/ARQ/function/aggregate#stdev_samp>`
- `<http://jena.apache.org/ARQ/function/aggregate#variance>`
- `<http://jena.apache.org/ARQ/function/aggregate#var_pop>`
- `<http://jena.apache.org/ARQ/function/aggregate#var_samp>`

As listed by Ontotext here: https://graphdb.ontotext.com/documentation/10.3/sparql-functions-reference.html?highlight=std#aggregation-function-extensions

Generally, the DISTINCT keyword can be used with these functions, but MySQL, MariaDB, DB2, and Oracle don't support it.

The new aggregate functions need to be added to the `CustomAggregateFunctionRegistry`, otherwise the SPARQL parser will not recognise them as aggregation functions. For this, the definition of factory classes that each provide the IRI of the corresponding aggregation function is required.

Also added test cases for the new functions to the lightweight test suits `DistinctInAggregate` and `BindTestWithFunctions`.


bcogrel (changes requested):
Hi @DRovara , very nice, thanks! 

Just one small improvement: could you please factorize the AVG, STD_x and VARIANCE_x SPARQL function symbols into a common abstract class? I guess most of the logic is the same for these three classes (typing, decomposition, etc.).

bcogrel (approved):

bcogrel:
Thanks a lot Damian!

PR #745, opened 2023-08-22
Fixed the implementation of the Redshift REGEXP_MATCH_3 function

DRovara:
Previously, we used the same implementation as postgres. This implemented RegEx flags like `"i"` by prepending the corresponding embedded options (e.g. `(?i)`) to the pattern string.

However, the Redshift RegEx dialect does not support embedded options. 
Fixed by using the function `REGEXP_INSTR` instead of the RegEx operator `~` and passing the flags as `parameter` flags to the function.


bcogrel (approved):

bcogrel:
Great, thanks Damian!

PR #746, opened 2023-08-23
Option to disable registering custom SPARQL aggregation functions

DRovara:
Option (`ontop.registerCustomSPARQLAggregateFunctions`) is included in `OntopKGQuerySettings` as it is accessed in `RDF4JQueryTranslatorImpl`.

When option is set to false (default true), custom aggregation functions (STDEV, VARIANCE...) will no longer be registered. This will make SPARQL queries using any of these functions fail during parsing.


bcogrel:
Thanks!

PR #753, opened 2023-09-06
Extract user, group and role information from HTTP headers

bcogrel:
This PR introduces a mechanism for extracting user, group and role information from HTTP headers and making it accessible through the following SQL functions:
 - `ontop_user()`
 - `ontop_contains_role_or_group('myRoleOrGroup')`

These functions can be used in SQL queries in the mapping and in lenses to implement access control.

It extracts information from the following HTTP headers: `x-user`, `x-roles` and `x-groups`.

By default, this information is not extracted. To enable, set `ontop.authorization=true`.


bcogrel:
Two more functions have been introduced in https://github.com/ontop/ontop/commit/528d41e6e5713386ef04ae33c4a81d1f0a0ee1c0 to distinguish roles from groups: `ontop_contains_role('myRole')` and `ontop_contains_group('myGroup')`

PR #755, opened 2023-09-07
Simplify Conjunctions and Disjunctions of Equalities

DRovara:
This pull request handles the issue of merging separate operands of boolean binary operations if they contain compatible equality checks.

We focus on equality checks of the form `<variable> = <constant>` or `<constant> = <variable>` only. Similar logic would also apply if we allow constants or (deterministic) function calls instead of `<variable>`, but the variable case is the most common and interesting. Similarly, the current approach might be extended to work with variables and function calls instead of `<constant>` as well, but this would make the task of cancelling contradictions much more complex, as `x = 'A' AND x = 'B'` is obviously a contradiction and can be cancelled, whereas `x = 'A' AND x = y` cannot be cancelled, and instead only implies `y = 'A'`.

### Strategy
The equality checks will be merged or cancelled by translating them into `IN` expressions which can be combined based on their arguments. Here, we define $x \in {1, 2, 3}$ as the function `IN(x, 1, 2, 3)`. The first term is called the _search term_ and the second is the _value list_.

We split the task into two smaller steps:
1) Transforming equality checks into `IN` expressions (e.g. `x = 'A'` ==> `IN(x, 'A')`) and transforming disjunctions of equalities into merged `IN` expressions (e.g. `x = 'A' or x = 'B'` ==> `IN(x, 'A', 'B')`)
2) Merging related `IN` expressions with each other.

Step (1) is trivial and can be completed in complexity `O(n)`
Step (2) is not as trivial and most likely NP-Hard. This pull request proposes a greedy heuristic using early stopping.

#### Merging Expressions.

Given a binary operation of the form `<expression1> * <expression2>` where * is `AND` or `OR` the task of _merging_ the expression involves taking advantage of simplification rules to combine `IN` terms from one expression with `IN` terms from the other expressions to minimize the total complexity of the query.

This can occur in one of three possible ways. We investigate the potential number of terms in the resulting merged expression:

___Both `expression1` and `expression2` are `IN` expressions___: This is the simplest case. If the expressions have different _search terms_, nothing can be done and we keep the binary operation as it is. Otherwise, we can create a combined `IN` expression where its _value list_ is the **SET UNION** for the operator `OR` and the __SET INTERSECTION__ for the operator `AND`.
In the worst case, the number of terms of the result will be the same as the original, and in the best case, it will go from 2 to 1.

___`expression1` is an `IN` expression and `expression2` is another boolean operation (or vice versa):___ In this case, the `IN` expression has to be added to the larger expression. If the outer boolean operator and the operator of `expression2` are the same, then we can simply add `expression1` to the list of operands in `expression2`. If the operators are different, we need to "_cross multiply_" `expression1` to each operand of `expression2`. This can be achieved by recursively calling the merging procedure for `<expression1> * <child_i>` for each operand of `expression2`, where `*` is _the same_ operator as in `<expression1> * <expression2>`. In the best case scenario (both operators are the same), the number of terms stays the same after merging (it may even decrease by 1, if we first try to see if `expression1` can cancel out with any operand of `expression2`). In the worst case, `expression1` has to be merged with each operand of `expression2` and is not compatible with any of them, doubling the total number of terms.

___Both `expression1` and `expression2` are further boolean  expressions:___ In this case, we first split `expression1` into its children. We then merge `<child_i> * <expression2>` for each child, and finally we merge all resulting expressions with each other. The number of resulting terms scales quadratically with the number of original terms.

Using these merging rules, we finally define the algorithm to simplify expressions as follows:
```
Start with the outermost boolean expression E = e_1 * e_2 * ... * e_n
current = e_1
for e_i in [e_2, ..., e_n]
	current = merge(current, e_i)
```

However, this algorithm makes the number of terms grow exponentially. Therefore, we implemented further optimisations:

___Merging with same operation___:
In the case `(x AND y) AND z`, where `x`, `y`, and `z` are `IN` expressions, we can always create a resulting formula `x AND y AND z` by the law of associativity. However, we can also check if `z` is "compatible" with either `x` or `y` (i.e. its `IN` call can be combined with one of theirs). In that case, instead of just adding `z` to the conjunction, we can combine it with its compatible neighbour, resulting in `x' AND y'` - decreasing the number of terms by 1. 

___Pre-determine compatibility___:
In cases (2) and (3) above, before "cross multiplying" the terms, we first make sure that both expressions have at least one `IN` expression with a shared _search term_. If this is not the case, we do not "cross multiply" and just leave the expression as it is. This decreases the likelihood of the total number of terms exploding, but the worst case is still exponential //TODO even better: factor out parts that are compatible

___Cancel unnecessary operands___:
In a conjunction `x AND y`, if `x` implies `y`, it is equivalent to just `x`. In a disjunction `x OR y`, if `x` implies `y`, it is equivalent to just `y`. Whenever we add a term to the full merge product, we can check if it will make any other terms obsolete and remove them.

___Early Stopping:___
As the number of terms can grow quadratically with each merge, adding an early stopping mechanism is a useful security tool. Here, we added the constant `MAX_ARITY` to `DisjunctionOfEqualitiesMergingSimplifierImpl`. That way, when we encounter a conjunction or disjunction during merging that has an arity larger than `MAX_ARITY`, we do not perform any further cross-multiplying. This prevents the terms from ever growing too large, but it also prevents larger / more complex expressions from being simplified.

___Greedy Ordering:___
Since we perform early stopping, it may make sense to merge the terms in such an order that the exponential explosion happens as late as possible. This way, we can attempt to merge the largest possible number of therms before being forced to stop. We achieve this by first sorting the terms that are to be merged, in the following order: `IN` expressions -> `AND/OR` expressions with same operator as parent -> `AND/OR` expressions with different operator from parent -> other expressions.
When just merging `IN` expressions, as mentioned above, the number of terms does not grow by much. Once the merged `IN` expressions become more complex boolean operations, each further `IN` expression will double the number of terms, and only at the very end we start adding complex expressions that will square the number of terms. 

### Discussion

The problem of simplifying boolean expressions in general is shown to be NP-Complete.
We can easily encode that problem into the more specialised problem of simplifying boolean expressions of equalities, so it should be NP-Hard as well. Therefore, finding a perfect solution is unlikely.
Without the early stopping condition, this approach should be able to simplify all expressions, therefore, a reasonable follow-up would be to decrease the rate at which the number of terms grows, so that the algorithm stops later. 

One idea for that would be to improve the ___Pre-determine compatibility___ strategy: For now, we only check if there is a compatible pair among both expressions. Another possible solution would be to find that compatible pair and factor it. That way, the number of terms should not increase at all in this situation. However, this also brings new problems: (1) how can we factor similarities out? The expressions might be complex, so this might be a complex task once again. (2) what if there are multiple compatible parts? (3) some factorings may only become "visible" after first merging with another `IN` term. Therefore, the order in which terms will be merged starts to become important, too. (4) factoring may not always be enough. In certain conditions, factoring can be blocked at some level. Now, we would need to verify, if it is still worth it to "cross-multiply" down until we reach the level where the factoring was being blocked to cancel that part out.
I have investigated this approach for a while without much success, but I also believe it could still be further investigated and improved on.

Alternatively, one might find a better way to order the terms before merging, so that we don't just increase the likelihood of finding the simplest merges but so that we can guarantee the simple merges are looked at first.


bcogrel (approved):

core/model/src/main/java/it/unibz/inf/ontop/model/term/functionsymbol/db/impl/DBInFunctionSymbolImpl.java
```diff
@@ -0,0 +1,87 @@
...
+
+import java.util.function.Function;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+public class DBInFunctionSymbolImpl extends DBBooleanFunctionSymbolImpl implements DBInFunctionSymbol {
+
+    protected DBInFunctionSymbolImpl(int arity, DBTermType rootDBTermType, DBTermType dbBooleanTermType) {
+        super("DB_IN_" + arity, IntStream.range(0, arity)
+                .mapToObj(i -> (TermType) rootDBTermType)
```
kontchakov, line 19: No need for the type cast: use .<TermType>mapToObj instead.

core/optimization/src/main/java/it/unibz/inf/ontop/iq/optimizer/impl/DisjunctionOfEqualitiesMergingSimplifierImpl.java
```diff
@@ -0,0 +1,474 @@
...
+        private boolean isImpliedBy(ImmutableFunctionalTerm determinant, ImmutableFunctionalTerm dependent) {
+            if(determinant.getFunctionSymbol() instanceof DBInFunctionSymbol && dependent.getFunctionSymbol() instanceof DBInFunctionSymbol) {
+                //IN1 --> IN2 <==> search(IN1) = search(IN2) AND terms(IN1) subsetof terms(IN2)
+                return determinant.getTerm(0).equals(dependent.getTerm(0))
+                        && Sets.difference(ImmutableSet.copyOf(determinant.getTerms()), ImmutableSet.copyOf(dependent.getTerms())).isEmpty();
+            }
+            if(determinant.getFunctionSymbol() instanceof DBInFunctionSymbol && isBooleanOperation(dependent)) {
+                if(dependent.getFunctionSymbol() instanceof DBAndFunctionSymbol) {
+                    return dependent.getTerms().stream()
+                            .allMatch(term -> (term instanceof ImmutableFunctionalTerm) && isImpliedBy(determinant, (ImmutableFunctionalTerm) term));
```
kontchakov, line 367: a second version of isImpliedBy could be applicable to ImmutableTerm, with the instanceof check and the type case to shorten these four lines

kontchakov (approved):
I really wonder whether the first argument of the new function can be removed from the list so that the numerous .skip operations can be avoided...

Also, in some DBs, using IN can be more efficient than a disjunction of equalities, so perhaps it makes sense to delay unpacking the INs until the SQL serializer (if that does not make other optimisations trickier).

core/model/src/main/java/it/unibz/inf/ontop/model/term/functionsymbol/db/impl/DBInFunctionSymbolImpl.java
DRovara: Done, thanks a lot!

core/optimization/src/main/java/it/unibz/inf/ontop/iq/optimizer/impl/DisjunctionOfEqualitiesMergingSimplifierImpl.java
DRovara: I believe this should now be done, too, if I understood you correctly.

bcogrel:
Amazing! Thanks a lot Damian!

PR #759, opened 2023-09-14
Propagate foreign keys from basic lenses to parents

DRovara:
Under certain conditions, foreign key constraints explicitly stated for basic lenses can be propagated to the lens' parent relation.

These conditions are:
- The referencing column is not modified by the lens.
- The foreign key does not point to the parent relation itself.

Also includes a set of test cases to test the feature together with UC propagation.

### DIFF
--- a/db/rdb/src/main/java/it/unibz/inf/ontop/dbschema/impl/LensMetadataProviderImpl.java
+++ b/db/rdb/src/main/java/it/unibz/inf/ontop/dbschema/impl/LensMetadataProviderImpl.java
@@ -182,9 +182,9 @@ public class LensMetadataProviderImpl implements LensMetadataProvider {
                 jsonLens.insertIntegrityConstraints((Lens) relation, baseRelations, metadataLookupForFK,
                         getDBParameters());
 
-                /* Propagate UCs to the parent relation.
-                If at least one UC was sent up this way, repeat this process from the parent.
-                Keep going until either no UC could be propagated up or the root element is reached.
+                /* Propagate UCs and KFs to the parent relation.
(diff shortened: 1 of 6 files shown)


bcogrel:
Motivating example from Ontopic Data Hub:
```sparql
PREFIX sosa: <http://www.w3.org/ns/sosa/>
PREFIX schema: <http://schema.org/>
PREFIX noi: <http://noi.example.org/ontology/odh#>
SELECT * WHERE {
 ?f schema:containedInPlace ?place .
 ?h a noi:HistoricalObservation ; sosa:hasFeatureOfInterest ?f . 
} LIMIT 10000
```

That gets translated into 
```sql
 SELECT v2."istat_code" AS "s_istat_code1m11", v1."station_id" AS "station_id1m139", v1."type_id" AS "type_id1m139", TO_JSON(v3."timestamp")#>>'{}' AS "v0"
FROM "intimev2"."measurement" v1, "intimev2"."station_municipalities" v2, "intimev2"."measurementhistory" v3
WHERE (v2."istat_code" IS NOT NULL AND v1."station_id" = v2."id" AND v1."station_id" = v3."station_id" AND v1."type_id" = v3."type_id")
LIMIT 10000
```

due to the use of the lens `lenses.measurement_station_municipalities_join`, which joins the tables `measurement` and `station_municipalities` together. The foreign key between `lenses.import.intimev2.measurementhistory` and `intimev2.measurement` on the column `station_id` and `type_id` is not used as all the lenses are at the same level and are unfolded at the same time.

bcogrel:
Great, thanks!

PR #760, opened 2023-09-14
Fixed implementation of Functional Dependency inference for Unions

DRovara:
We now use a different strategy to infer FDs of unions. This now works even if the set of dependents or determinants are not equal.

For instance, if `relation1` and `relation2` have the following dependencies, respectively:
`(a, b) -> (x, y)`, `(a, b, c) -> (x)`

Then we can infer the FD of the union (assuming at least one column is disjoint):
`(a, b, c) -> (x)`


bcogrel (approved):

bcogrel:
Thanks Damian!

PR #771, opened 2023-10-22
No default schema in MS SQL Server for Ontop 5.1

kontchakov:
The JDBC driver for MS SQL Server in certain settings returns `null` instead of the default schema name. A special implementation of metadata extraction is provided to account for this case.


bcogrel:
Thanks Roman!

PR #775, opened 2023-11-03
Minor fix of geof:buffer implementation in PostGIS

apano-on:
In PostGIS, the GeoSPARQL function geof:Buffer used with argument uom:meter and a geometry on a coordinate system with units in degrees (e.g. EPSG:4326), performed a CAST to GEOGRAPHY to yield correct results (see workaround also [here](https://gis.stackexchange.com/questions/387144/st-buffer-in-meters-from-4326-point-table)). However, the serialization of the argument to WKT before the CAST was not removed. This meant a geospatial index for ST_ASTEXT(geom)::GEOGRAPHY in the SQL database could not be added and the performance of geof:Buffer was quite poor in PostGIS.

This pull request:
- Unwraps the unnecessary cast to TEXT. A geospatial index for geom::GEOGRAPHY can be added and the performance of geof:Buffer in PostGIS for large datasets improves substantially
- Adds some PostGIS tests to the Github CI
- Drops previously redundant PostGIS tests in Gitlab docker suite
- Adds a missing geospatial transform DBFunctionSymbol, which was not getting simplified for some queries


bcogrel:
Thanks Albulen!

PR #779, opened 2023-12-14
Encode to utf8 the mysql datetime conversion

skomlaebri:
Mysql datetime conversion requires a cast with character set utf8 to avoid collation error.
All strings in MySQL are already cast to utf8 as visible here:
https://github.com/ontop/ontop/blob/2606d30b83fb01269525d1342947e13bc503a04c/db/rdb/src/main/java/it/unibz/inf/ontop/model/type/impl/MySQLDBTypeFactory.java#L49


bcogrel (approved):

bcogrel:
Thanks Sarah!

Issue #791, opened 2024-02-22
In objectMap two distinct SQL queries without rr:joinCondition must result in cross-joins


samdus:
### Description

We may use a cross-join between two logical tables when defining an R2RML mapping for an object property. The way to achieve this is to define a rr:objectMap without specifying a rr:joinCondition. Moreover, R2RML [specifies](https://www.w3.org/TR/r2rml/#foreign-key) :

> A referencing object map is represented by a resource that **may** have one or more rr:joinCondition properties, whose values must be join conditions.

Therefore, the join conditions are optional, and Ontop should support object maps without rr:joinCondition. Unfortunately, R2RMLToSQLPPTriplesMapConverter will throw an exception if the objectMap doesn't contain a rr:joinCondition and the two SQL queries are different.

### Steps to Reproduce

1. Create an object property mapping object property without specifying a rr:joinCondition
3. Load the query engine with the mapping
4. The exception is thrown before the load is completed.

**Expected behavior:** [What you expect to see]
The query engine loads successfully. The query generated for the object property contains a cross-join if the logicalTable refers to a table. The query generated for the object property contains the added constant if the logicalTable refers to a constant.

**Actual behavior:** [What actually happens]
The exception is thrown: 
````
java.lang.IllegalArgumentException: No rr:joinCondition, but the two SQL queries are distinct: [...]
	at it.unibz.inf.ontop.spec.mapping.parser.impl.R2RMLToSQLPPTriplesMapConverter.getRefSQLPPTriplesMaps(R2RMLToSQLPPTriplesMapConverter.java:203)
````

**Reproduces how often:** [Does it happen every time?]
Reproduces every time.

### Attached material

An example mapping can be found here: https://github.com/samdus/ontop/tree/bugfix/emptyJoinConditionSupport/test/docker-tests/src/test/resources/issue791

### Versions

Current "version5" branch (as of 2024-02-22, commit 6216b518f37425a670ec29f4a02fd409e1b8c9f8).

kontchakov:
Hi Samuel 

Thanks for the proposed fix. The [R2RML specification](https://www.w3.org/TR/r2rml/#foreign-key), however, says 

[...]

PR #792, opened 2024-02-22
Bugfix: support of object maps without join conditions

samdus:
Added support of rr:objectMap without rr:joinCondition.

Fixes ontop/ontop#791


samdus:
Since the proposed modification does not respect the current R2RML specification, I close the PR. For more detail, please look at the discussion in ontop/ontop#791.

Issue #803, opened 2024-04-27
Left Join Failure


jp3477:
### Description

I am attempting to create a mapping on top of an MSSQL database. In my mapping's source query, I have a JOIN that succeeds when it is an INNER JOIN, but fails when it is an OUTER JOIN (LEFT & RIGHT). Some of the fields extracted from the JOIN clause are fed into placeholders within the target triples. Does Ontop support OUTER JOINs?

### Steps to Reproduce

1. Add mapping entry to `ont.obda` file with LEFT JOIN in source query
2. Start ontop client `ontop-cli-5.1.2/ontop endpoint --ontology=ont.ttl --mapping=ont.obda --properties==ont.properties`
3. Add mapping entry to .obda file with INNER JOIN in source query
4. Start ontop client `ontop-cli-5.1.2/ontop endpoint --ontology=ont.ttl --mapping=ont.obda --properties==ont.properties`

#### Mapping entry
```
mappingId	Observation.valueQuantity
target		:Observation/{measurement_id} fhir:value :Observation/{measurement_id}/valueQuantity . :Observation/{measurement_id}/valueQuantity a fhir:Quantity ; fhir:value :Observation/{measurement_id}/valueQuantity/value ; fhir:unit :Observation/{measurement_id}/valueQuantity/unit ; fhir:codeP :Observation/{measurement_id}/valueQuantity/code . :Observation/{measurement_id}/valueQuantity/value a fhir:decimal ; fhir:v {value}^^xsd:float . :Observation/{measurement_id}/valueQuantity/unit a fhir:string ; fhir:v {unit}^^xsd:string . :Observation/{measurement_id}/valueQuantity/code a fhir:code ; fhir:v {code}^^xsd:string . 
source		SELECT
                    measurement_id,
                    m.value_as_number [value],
                    unit_c.concept_code unit,
                    unit_c.concept_code code
                FROM
                    measurement m
                    LEFT JOIN concept unit_c ON unit_c.concept_id = m.unit_concept_id
                    AND unit_c.concept_id != 0
                WHERE
                    m.value_as_number IS NOT NULL
```

**Expected behavior:**

It is expected that running the ontop client will run successfully after this mapping is added to the `ont.obda` file.

**Actual behavior:**

The ontop client fails to load. Here is a relevant excerpt:

```
Caused by: it.unibz.inf.ontop.exception.InvalidMappingSourceQueriesException: The placeholder(s) value in the target do(es) not occur in source query of the mapping assertion
[...]

PR #805, opened 2024-04-29
Fix for issue #803

kontchakov:
MS SQL Server identifiers are not case-sensitive (even if quoted).

The keyword `AS` for introducing column aliases is optional. The updated approximate column name extraction now allows that.


bcogrel:
Thanks Roman!

PR #819, opened 2024-07-31
GH-818 upgrade to RDF4J 5.0.1

### LINKED ISSUE #818: Migrate to RDF4J 5.0.1
### Description

Ontop currently uses RDF4J 4.3.10. The RDF4J 4.x.x version will not receive any further bug fixes or improvements. Migrating to RDF4J 5.0.1 insures that Ontop stays on a major version that will continue to be maintained.

### Versions

Ontop 5.1.2.

### Additional Information

RDF4J 5 contains breaking changes when compared to 4.3.x. A general question is whether upgrading to RDF4J 5 can be done in Ontop 5.2 or if it would have to be scheduled for Ontop 6.

hmottestad:
Fixes #818

Upgrade from RDF4J 4.3.10 to RDF4J 5.0.1.


hmottestad:
OWLAPI also needs to be upgraded to RDF4J 5.

hmottestad:
@bcogrel @ghxiao

What are your thoughts on upgrading to RDF4J 5.0.1? 

@reckart and I have managed to upgrade OWLAPI and once @ignazio1977 releases the upgrade then this PR here should be ready.

bcogrel:
Thanks @hmottestad for this PR!

To me, it makes sense doing such upgrade once the new version of OWLAPI is out.
In the meantime, is the snapshot version of OWLAPI 5.5.1 published somewhere, so as to have the pipeline running?

ignazio1977:
> owlapi snapshots builds

They *can* be published via Sonatype but that doesn't happen very often. If I remember correctly snapshots are deleted after a few days.

It would be a manual process either way.

bcogrel:
Thanks for the info. I was just asking in case the snapshots were already available so we can already test.

I don't think it is worth the effort, so I suggest just waiting for a new release of OWLAPI.

hmottestad:
> > owlapi snapshots builds
> 
> They _can_ be published via Sonatype but that doesn't happen very often. If I remember correctly snapshots are deleted after a few days.
> 
> It would be a manual process either way.

RDF4J  has a nightly snapshot build that is published to the oss sonatype snapshots repo: https://github.com/eclipse-rdf4j/rdf4j/blob/eab72726d7814fe73476f78477adf27e4ff67155/pom.xml#L57

Not that you need to set up anything similar. Just to show how we've solved it. I don't think it's used very often though.

ignazio1977:
Owlapi used to have a travis job that could have included deploying snapshots, but travis stopped offering free builds a while ago and we don't have any funding to tap into :-) so it fell by the wayside. Not much activity on the owlapi repositories either, these days.

hmottestad:
RDF4J is lucky to be part of the Eclipse organisation. Someone at Eclipse is in charge of maintaining several Jenkins environments and we get to use one for free. And when there is downtime we can post an issue in their Gitlab and it's up an running the next day :) 

The integration between GitHub  and Jenkins was never that great and we only use Jenkins for nightly snapshot builds and when we publish a new release of RDF4J. For anything related to build stability, tests and PRs we use GitHub Actions, which is free :D Though you never know when Microsoft will be happy with their market share and start charging for that too :P

hmottestad:
@ignazio1977 Do you have a plan when you will be releasing the next OWLAPI version?

ignazio1977:
@hmottestad I'll add in the bom and release, I think. I'll have a quick check for library updates as well. In terms of deadlines, perhaps a couple of weeks.

ignazio1977:
@hmottestad version 5.5.1 just released

hmottestad:
Thank you @ignazio1977 !

@bcogrel Would you like to take a look and run the tests?

bcogrel:
Thanks for releasing OWLAPI and for updating the branch! Tests are passing.

We need to understand to which version attaching these changes, as we released 5.2.0 a few weeks ago.
Let us discuss it within the Ontop team next week.

bcogrel:
Our plan is to release it soon as 5.3.0. We need first to test the Ontop Protégé plugin, as the co-existence of two versions of OWLAPI (4 and 5) there is a bit brittle.

Issue #822, opened 2024-08-23
Swapping in construction nodes results in invalid queries


samdus:
### Description

During the saturating process, Ontop reuses query definitions defined for an object property to define any inverseOf object property. It usually works as intended, but I ran across a specific use case where the generated query is not bound correctly. 

The method it.unibz.inf.ontop.spec.mapping.transformer.impl.TMappingSaturatorImpl.MappingAssertionConstructionNodeTransformer#updateConstructionNodeIri will generate a ConstructionNode that simply swaps the subject and object of the child tree. However, as I understand the definition of ConstructionNode, it is illegal to have a variable in the domain and range of the Substitution. In my use case, it causes the childVariables to be empty. Therefore, every variable is considered as not required and is removed by it.unibz.inf.ontop.iq.node.normalization.impl.NotRequiredVariableRemoverImpl.

I'm not so sure that my proposition goes in the spirit of the rest of the code base, but I would change the ConstructionNode constructor to permit the specific case when the construction node is used to make a swap. I made a [Pull Request](https://github.com/ontop/ontop/pull/823) for an example of this proposition.

### Steps to Reproduce

1. I defined [a test in my fork](https://github.com/samdus/ontop/blob/48f32e2e20812ebb3f80d5d5c191038f5e0792f4/core/model/src/test/java/it/unibz/inf/ontop/iq/NotRequiredVariableRemoverTest.java) to illustrate the behaviour.
2. An example of Tree that are currently invalid is :
```
CONSTRUCT [s, p, o] [s/o, p/<http://example.org/iri>, o/s]
   UNION [s, o]
      CONSTRUCT [s, o] [s/RDF(http://example.org/dom1/{}(v0),IRI), o/RDF(http://example.org/range1/{}(v1),IRI)]
         VALUES [v0, v1] ("1"^^STRING,"2"^^STRING)
      CONSTRUCT [s, o] [s/RDF(http://example.org/dom2/{}(v2),IRI), o/RDF(http://example.org/range2/{}(v3),IRI)]
         VALUES [v2, v3] ("3"^^STRING,"4"^^STRING)
```

**Expected behavior:** [What you expect to see]

The tree should be valid and, therefore, could be optimized as follows:
```
CONSTRUCT [s, p, o] [p/<http://example.org/iri>, s/RDF(v1,IRI), o/RDF(v0,IRI)]
[...]

PR #823, opened 2024-08-23
Swaps could be valid in Construction Node 

samdus:
During the saturating process, Ontop reuses query definitions defined for an object property to define any inverseOf object property. It usually works as intended, but I ran across a specific use case where the generated query is not bound correctly. 

The method it.unibz.inf.ontop.spec.mapping.transformer.impl.TMappingSaturatorImpl.MappingAssertionConstructionNodeTransformer#updateConstructionNodeIri will generate a ConstructionNode that simply swaps the subject and object of the child tree. However, as I understand the definition of ConstructionNode, it is illegal to have a variable in the domain and range of the Substitution. In my use case, it causes the childVariables to be empty. Therefore, every variable is considered as not required and is removed by it.unibz.inf.ontop.iq.node.normalization.impl.NotRequiredVariableRemoverImpl.

I'm not so sure that my proposition goes in the spirit of the rest of the code base, but I would change the ConstructionNode constructor in order to permit the specific case when the construction node is used to make a swap.

Fixes https://github.com/ontop/ontop/issues/822

### DIFF
--- a/core/model/src/main/java/it/unibz/inf/ontop/iq/node/impl/ConstructionNodeImpl.java
+++ b/core/model/src/main/java/it/unibz/inf/ontop/iq/node/impl/ConstructionNodeImpl.java
@@ -62,9 +62,8 @@ public class ConstructionNodeImpl extends ExtendedProjectionNodeImpl implements 
 
         // only the variables that are also used in the bindings for the child of the construction node
         this.childVariables = Sets.difference(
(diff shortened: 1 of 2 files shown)


samdus:
As stated in https://github.com/ontop/ontop/issues/822, it is not desirable to permit swaps in Construction Node.

Issue #836, opened 2025-01-19
`geof:distance` with trino connector


justin2004:
ontop is generating SQL for SPARQL with this BIND
```
bind(geof:distance(?wkt,?wkt1,uom:metre) as ?dist)
```

but one of the functions in the SQL isn't a native trino function:
```
io.trino.spi.TrinoException: line 2:426: Function 'st_distancesphere' not registered
```

but it does look like trino has a function that would work:
https://trino.io/docs/current/functions/geospatial.html#ST_Distance

is it possible for ontop to emit SQL with that `ST_Distance` function instead of `st_distancesphere` when the trino connector is used?

apano-on:
That should indeed be correct and I quickly tested it. You would just need to specify ```distance``` instead of ```distancesphere``` for this case:
```
@Override
    public DBFunctionSymbol getDBSTDistanceSphere() {
        return getRegularDBFunctionSymbol(ST_DISTANCE, 2);
    }
```
Based on how ST_DISTANCE is documented in Trino, it seems to be identical to ST_DISTANCESPHERE in PostGIS (which was the default used in this case). Only the unit ```uom:metre``` seems supported for distance in Trino, cases where input geometry coordinates are in degrees (e.g. EPSG:4326) and output unit is metres.

justin2004:
hey @bcogrel , any update on this?

bcogrel:
Hi @justin2004 , my colleague @ManuelaCorte will have a look at it.

justin2004:
thanks @bcogrel . if @ManuelaCorte needs an environment to test against i've described a docker image with commands to load Geometry data [here](https://github.com/trinodb/trino/issues/24961#issue-2840408539).

justin2004:
i setup an environment that reproduces this error here:
https://github.com/justin2004/ontop-jig?tab=readme-ov-file#does-not-work-1

bcogrel:
Fixed by #845. Thanks Manuela!

PR #845, opened 2025-02-25
ST_Distance() function for Trino

ManuelaCorte:
The geospatial function `ST_Distance()` for Trino does not work as the default used in Ontop (which is taken from PostGIS) but instead works similar to the `ST_DistanceSphere()` function. Moreover the `ST_Distance()` function in Trino returns the 2-dimensional cartesian minimum distance or the great-circle distance based on whether the arguments are geometries or geographies respectively.

This PR adds support for the `ST_Distance()` function in Trino and fixes issue #836


bcogrel (approved):

justin2004:
great!

PR #849, opened 2025-03-04
allow more trino functions in lens expressions

justin2004:
allow lens expressions (for trino) such as:
```
st_astext(ST_GeometryFromText(format('Point(%s %s)', json_extract(geometry, '$.coordinates[0]'), json_extract(geometry, '$.coordinates[1]'))))
```

and the resultant values to be used as arguments to `geof:distance`


justin2004:
@apano-on , can you tell what is failing here?
https://github.com/ontop/ontop/actions/runs/13662343115/job/38318282308?pr=849

apano-on:
2 MySQL tests are failing, but they actually pass locally (I checked) so there is likely a Github CI consistency issue unrelated to this PR.

By the way, your observation that the db term type should be ```dbGeometryType``` rather than ```dbStringType``` is correct, and we are working to fix a few geo-related DB function symbols with a similar issue.

apano-on:
Correction: Testing your branch (I apologize for testing the incorrect local branch), the MySQL tests failing are:
- [testFlattenJsonPossiblyNull](https://github.com/ontop/ontop/blob/version5/test/lightweight-tests/src/test/java/it/unibz/inf/ontop/docker/lightweight/AbstractNestedDataTest.java#L141)
- [testSpo](https://github.com/ontop/ontop/blob/version5/test/lightweight-tests/src/test/java/it/unibz/inf/ontop/docker/lightweight/AbstractNestedDataTest.java#L178)

which are returning an invalid ```RDF(NULL, xsd:integer)``` due to ```JSON_EXTRACT``` which is also used in MySQL (e.g. [here](https://github.com/ontop/ontop/blob/version5/test/lightweight-tests/src/test/resources/nested/mysql/nested-lenses.json#L90) flatten lenses)

Adding explicitly JSON_EXTRACT as a DBFunctionSymbol might be causing an issue with the nested type support in MySQL https://ontop-vkg.org/guide/databases/mysql.html#nested-type-support impact by the explicit definition of

justin2004:
maybe i should move JSON_EXTRACT into TrinoDBFunctionSymbolFactory? i am able to use that function in lens expressions for trino.

justin2004:
hey @apano-on , any idea why i can't trigger the CI tests? i made a change and i'd like to see if it passes the mysql tests now.

justin2004:
hey @bcogrel , are CI tests disabled?

bcogrel:
Here is the CI run: https://github.com/ontop/ontop/actions/runs/13662343115

bcogrel:
@justin2004 we indeed experienced some issues with the CI pipeline recently, which should be fixed now. 
Could you please merge `version5` into your branch?

justin2004:
hey @bcogrel , i merged `version5` into this branch. CI tests passed locally.

bcogrel:
Tests are passing also on Ontop's CI.

@justin2004, may I ask to register these functions only in the class `TrinoDBFunctionSymbolFactory` and keep the class `AbstractSQLDBFunctionSymbolFactory` as it was?

Afterwards I think we are good to go.

Thanks!

justin2004:
@bcogrel , but the functions `FORMAT` and `ST_GEOMETRYFROMTEXT` are not Trino specific. I thought AbstractSQLDBFunctionSymbolFactory was the place for functions that are in several DB implementations?

bcogrel:
Ok, you are right. Let me merge your branch then. Thanks!

PR #853, opened 2025-03-25
Partial support for FILTER NOT EXISTS

ManuelaCorte:
The `FILTER NOT EXISTS` clause is currently not supported since from an algebraic point of view it is evaluated in a top-down manner, while Ontop evaluates queries with a bottom-up approach.

 However, in the majority of cases, the `FILTER NOT EXISTS` clause can be replaced by a `MINUS` clause which instead can be evaluated in a bottom-up manner. The differences between the two are discussed in the [SPARQL 1.1 specification](https://www.w3.org/TR/sparql11-query/#neg-notexists-minus) but in general, the two are not equivalent when:
- No non-nullable variables are shared between the left and right side of the `FILTER NOT EXISTS` clause.
- Inner filters  in the `FILTER NOT EXISTS` involving variables not in its scope. For example, the following query
    ```sparql
    PREFIX : <http://example.com/>
    SELECT * WHERE {
            ?x :p ?n
            FILTER NOT EXISTS {
                    ?x :q ?m .
                    FILTER(?n = ?m)
            }
    }
    ```
    cannot be translated into a `MINUS` clause since in a bottom-up parsing, the filter `?n = ?m` cannot be evaluated since `?n` is not bound.

The `MINUS` is translated in an `IQTree` using a `LEFT JOIN` followed by a `FILTER` checking whether a variable unique to the right side of the join is `NULL`. Additionally, it is also required that at least one of the shared variables between the left and right child of the left join is non nullable.

For example, the following query:
```sparql
PREFIX : <http://www.example.org/>
SELECT ?person
WHERE {
  ?person a :Person .
  FILTER NOT EXISTS { ?person :firstName ?fname }
}
```
is translated into the following `IQTree`:
```
CONSTRUCT [person] []
   FILTER IS_NULL(fname)
      LJ
         CONSTRUCT [person] [person/RDF(http://person.example.org/person/{}(INTEGERToTEXT(id1m1)),IRI)]
            EXTENSIONAL "person"(0:id1m1)
         CONSTRUCT [person, fname] [person/RDF(http://person.example.org/person/{}(INTEGERToTEXT(id1m2)),IRI), fname/RDF(CHARACTER VARYINGToTEXT(firstName1m2),xsd:string)]
            FILTER IS_NOT_NULL(firstName1m2)
               EXTENSIONAL "person"(0:id1m2,1:firstName1m2)
```

### DIFF
--- a/binding/rdf4j/src/test/java/it/unibz/inf/ontop/rdf4j/repository/NotExistsTest.java
+++ b/binding/rdf4j/src/test/java/it/unibz/inf/ontop/rdf4j/repository/NotExistsTest.java
@@ -1,0 +1,119 @@
+package it.unibz.inf.ontop.rdf4j.repository;
+
+import com.google.common.collect.ImmutableSet;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Ignore;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.sql.SQLException;
+
+import static org.junit.Assert.assertEquals;
+
+public class NotExistsTest extends AbstractRDF4JTest {
+    private static final String OBDA_FILE = "/person/person_not_exists.obda";
+    private static final String SQL_SCRIPT = "/person/person_not_exists.sql";
+
+    @BeforeClass
+    public static void before() throws IOException, SQLException {
+        initOBDA(SQL_SCRIPT, OBDA_FILE);
+    }
+
+    @AfterClass
+    public static void after() throws SQLException {
+        release();
+    }
+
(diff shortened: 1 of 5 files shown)


bcogrel:
Note for a future improvement (not for this PR):

```sparql
PREFIX : <http://example.com/>
SELECT * WHERE {
        ?x :p ?n
        FILTER NOT EXISTS {
                ?x :q ?m .
                FILTER(?n = ?m)
        }
}
```
could be supported by adding `?n = ?m` as a joining condition of the left-join encoding the minus.

Issue #854, opened 2025-03-28
DuckDB - attach to data sources


larsw:
Hi,

 I've looked through the documentation, but could not find anything regarding attaching data sources in DuckDB - is there any way to run an "init"-script of some sort to handle this scenario?

bcogrel:
Hi @larsw, unfortunately not at the moment. But that's would be a nice feature to add for DuckDB. 
Thanks for suggesting it.

bcogrel:
Implemented by https://github.com/ontop/ontop/pull/856

PR #856, opened 2025-04-07
Database initialization script

ManuelaCorte:
This PR introduces a new setting to the Ontop properties file, `ontop.dbInitializationScript`, which allows users to add a script that will be executed each time a new connection is established to the database.

Since initialization scripts are usually quite small, the script is specified as a string in the properties file instead of as a separate file. For example, the following properties file creates an in-memory DuckDB database with a table and some initial data:
```
jdbc.url = jdbc:duckdb:
jdbc.driver = org.duckdb.DuckDBDriver
ontop.dbInitializationScript = CREATE TABLE person ( \
id TEXT PRIMARY KEY, \
name TEXT NOT NULL, \
age INT ); \
\
INSERT INTO person VALUES('1', 'Anna', 30); \
INSERT INTO person VALUES('2', 'Bob', 25); \
INSERT INTO person VALUES('3', 'Charlie', NULL);
```

It is important to note that Ontop may open multiple connections to the database, and the initialization script will be executed on each of them. This means that if you perform operations that are not idempotent (e.g., creating a table), you should ensure that the script can be executed multiple times without causing errors. For example, you can use `CREATE TABLE IF NOT EXISTS` to avoid errors when the table already exists. Another possible solution for in-memory databases such as DuckDB and H2 is to avoid specifying a database name in the JDBC URL like in the example above, which will create a new database on each connection.


bcogrel (approved):

bcogrel:
Thanks Manuela!

Closes https://github.com/ontop/ontop/issues/854 .

PR #861, opened 2025-04-28
Partial Implementation for SPARQL EXISTS

ManuelaCorte:
This PR implements a partial solution for the SPARQL `EXISTS` clause for cases where it can be evaluated in a bottom-up fashion, similar to the approach used in #853 for the `FILTER NOT EXISTS` case.

The `EXISTS` clause is translated using a `LEFT JOIN`, where the right child represents the subquery of the `EXISTS` clause. In addition to containing the `IQTree` for the EXISTS subquery, the right child ensures consistent cardinality by applying a `DISTINCT` operator to the variables that appear in the right child also present in the left child.

To evaluate whether the graph pattern of the `EXISTS` clause has been matched, an additional variable (`prov/"ontop-provenance-constant"`) is projected in the right child of the `LEFT JOIN`. This variable tracks whether the query pattern in the right subtree has been matched in the join. It's then used to create the expression that represents the result of the `EXISTS` clause evaluation using the bound function: `SP_BOUND(RDF(prov,xsd:string))`.

For example, given the following query:

```sparql
PREFIX  : <http://person.example.org/>

SELECT ?person ?hasNick WHERE {
   ?person a :Person .
   BIND(EXISTS { ?person :nickname ?nick } AS ?hasNick)
}
```

It will be translated into the following `IQTree`:

```
CONSTRUCT [person, hasNick] [hasNick/SP_BOUND(RDF(prov,xsd:string))]
   CONSTRUCT [person, prov] []
      LJ AND2(STRICT_EQ2(person,person0),STRICT_EQ2(person,person0))
         INTENSIONAL triple(person, <http://www.w3.org/1999/02/22-rdf-syntax-ns#type>, <http://person.example.org/Person>)
         DISTINCT
            CONSTRUCT [prov, person0] [prov/"ontop-provenance-constant"^^TEXT]
               INTENSIONAL triple(person0, <http://person.example.org/nickname>, nick)
```

In the previous example the `DISTINCT` operator is necessary when a person can have multiple nicknames. Consider these sample triples:

```
<http://person.example.org/1> a <http://person.example.org/Person> .
<http://person.example.org/1> <http://person.example.org/nickname> "Nick1" .
<http://person.example.org/1> <http://person.example.org/nickname> "Nick2" .
<http://person.example.org/2> a <http://person.example.org/Person> .
<http://person.example.org/2> <http://person.example.org/nickname> "Nick3" .
<http://person.example.org/3> a <http://person.example.org/Person> .
```

The correct result of the query would be:

| person                        | hasNick |
| ----------------------------- | ------- |
| <http://person.example.org/1> | true    |
| <http://person.example.org/2> | true    |
| <http://person.example.org/3> | false   |

Without the `DISTINCT` operator, each nickname would generate a separate row, resulting in:

| person                        | hasNick |
| ----------------------------- | ------- |
| <http://person.example.org/1> | true    |
| <http://person.example.org/1> | true    |
| <http://person.example.org/2> | true    |
| <http://person.example.org/3> | false   |


core/kg-query/src/main/java/it/unibz/inf/ontop/query/translation/impl/RDF4JValueExprTranslator.java
```diff
@@ -81,118 +90,150 @@ public ImmutableTerm getTerm(ValueExpr expr) {
...
         }
         if (expr instanceof Not) {
-            return getFunctionalTerm(XPathFunction.NOT.getIRIString(), convertToXsdBooleanTerm(term));
+            ImmutableFunctionalTerm notTerm = getFunctionalTerm(XPathFunction.NOT.getIRIString(), convertToXsdBooleanTerm(extendedTerm.getTerm()));
+            return new ExtendedTerm(notTerm, extendedTerm.getExistsMap());
         }
         if (expr instanceof IsNumeric) {
             // isNumeric (Sec 17.4.2.4) for checking whether the argument is a numeric value
-            return getFunctionalTerm(SPARQL.IS_NUMERIC, term);
+            return new ExtendedTerm(getFunctionalTerm(SPARQL.IS_NUMERIC, extendedTerm.getTerm()), extendedTerm.getExistsMap());
```
bcogrel, line 216: Please create a method to reduce the boilerplate for functions

core/kg-query/src/main/java/it/unibz/inf/ontop/query/translation/impl/RDF4JTupleExprTranslator.java
```diff
@@ -713,12 +821,20 @@ private TranslationResult translate(Extension node) throws OntopInvalidKGQueryEx
...
         ImmutableSet<Variable> childVars = childTranslation.iqTree.getVariables();
-        ImmutableList<Substitution<ImmutableTerm>> mergedVarDefs = getVarDefs(node.getElements(), childVars);
+        ValueExpressionResult<ImmutableList<Substitution<ImmutableTerm>>> varDefsResult = getVarDefs(node.getElements(), childVars);
 
+        ImmutableList<Substitution<ImmutableTerm>> mergedVarDefs = varDefsResult.getResult();
         if (mergedVarDefs.isEmpty()) {
             return childTranslation;
         }
 
+        if (varDefsResult.getExistsMap().size() > 1) {
```
bcogrel, line 831: Restriction to be relaxed

core/kg-query/src/main/java/it/unibz/inf/ontop/query/translation/impl/RDF4JValueExprTranslator.java
```diff
@@ -396,4 +465,28 @@ public VariableOrGroundTerm translateRDF4JVar(Var v, boolean leafNode) {
                 : Optional.ofNullable(externalBindings.get(var))
                         .orElseGet(termFactory::getNullConstant);
     }
+
+    public static class ExtendedTerm {
+        private final ImmutableTerm term;
+        private final ImmutableMap<Variable, ValueExpr> existsMap ;
```
bcogrel, line 471: Change type to `ImmutableMap<Variable, Exists>` ? I have seen many unchecked casts in the code.

core/kg-query/src/main/java/it/unibz/inf/ontop/query/translation/impl/RDF4JTupleExprTranslator.java
```diff
@@ -202,23 +197,37 @@ private TranslationResult translate(Group group) throws OntopInvalidKGQueryExcep
...
             throw new Sparql2IqConversionException("Unexpected parsed SPARQL query: nested complex projections appear " +
                     "within an RDF4J Group node: " + group);
         }
-        ImmutableSet<Variable> childVariables = child.iqTree.getVariables();
 
+        IQTree childTree;
+        if (varDefsResult.existsMap.isEmpty()) {
+            childTree = child.iqTree;
+        } else {
+            if (varDefsResult.getExistsMap().size() > 1) {
```
bcogrel, line 213: Duplicated logic. Please try to factorize.

core/kg-query/src/main/java/it/unibz/inf/ontop/query/translation/impl/RDF4JTupleExprTranslator.java
```diff
@@ -229,38 +238,69 @@ private TranslationResult translate(Group group) throws OntopInvalidKGQueryExcep
...
+                .filter(Optional::isPresent)
+                .map(Optional::get)
+                .map(RDF4JValueExprTranslator.ExtendedTerm::getExistsMap)
+                .filter(map -> !map.isEmpty())
+                .collect(ImmutableCollectors.toList());
+        if (existsMaps.size() > 1) {
+            throw new OntopUnsupportedKGQueryException("Multiple EXISTS in the same operator are not supported: " + order);
+        }
+
+        IQTree childTree;
```
bcogrel, line 287: Duplicated logic. Please try to factorize.

core/kg-query/src/main/java/it/unibz/inf/ontop/query/translation/impl/RDF4JTupleExprTranslator.java
```diff
@@ -514,7 +624,7 @@ private TranslationResult translateJoinLikeNode(BinaryTupleOperator join) throws
         if (join instanceof LeftJoin) {
             Sets.SetView<Variable> variables = Sets.union(leftTranslation.iqTree.getVariables(), rightTranslation.iqTree.getVariables());
             Optional<ImmutableExpression> filterExpression = Optional.ofNullable(((LeftJoin) join).getCondition())
-                    .map(c -> topSubstitution.apply(getFilterExpression(c, variables)));
```
bcogrel, line 517: Why ignoring the EXISTS map?

bcogrel: At the least we should throw an unchecked exception if we are in a case we don't support

core/kg-query/src/main/java/it/unibz/inf/ontop/query/translation/impl/RDF4JTupleExprTranslator.java
```diff
@@ -219,38 +222,73 @@ private TranslationResult translate(Group group) throws OntopInvalidKGQueryExcep
...
+        IQTree childTree;
+        if (existsMaps.isEmpty()) {
+            childTree = child.iqTree;
+        } else {
+            ImmutableList<IQTree> orderElementsExistsSubtrees = existsMaps.stream()
+                    .map(e -> {
+                        try {
+                            return translateExists(e, child);
+                        } catch (OntopUnsupportedKGQueryException | OntopInvalidKGQueryException ex) {
+                            throw new RuntimeException(ex);
```
bcogrel, line 277: Please do not wrap the exception into a RuntimeException. It is important we preserve the checked OntopUnsupportedKGQueryException. Use a for loop instead.

bcogrel (approved):

bcogrel:
Thanks a lot Manuela!

PR #862, opened 2025-05-02
IQTree and QueryNode code cleanup

kontchakov:
Reduced the four visitor-style method sets to a single one, for`IQVisitor`:`IQTreeVisitingTransformer` now extends `IQVisitor`; the extended `IQTreeVisitingTransformer` has been eliminated (only a `VariableGenerator` additional parameter was used and that never changed between method calls); finally, the visitor on `QueryNode` was not required - the double dispatch was done via `IQTree` anyway.

Other small improvements: uniform caching in `AbstractCompositeIQTree`; reduced code duplication in `JsonLens`; immutable data structures in `BasicFlattenLifter`.


bcogrel (approved):
It looks good. Thanks Roman!

Issue #858, opened 2025-04-09
Arity violation: FALSE_OR_NULL17 was expecting 17, not 5


justin2004:
ontop 5.3.0

this mapping file:
```
[PrefixDeclaration]
rdf:        http://www.w3.org/1999/02/22-rdf-syntax-ns#
rdfs:       http://www.w3.org/2000/01/rdf-schema#
owl:        http://www.w3.org/2002/07/owl#
xsd:        http://www.w3.org/2001/XMLSchema#
obda:       https://w3id.org/obda/vocabulary#

[MappingDeclaration] @collection [[

mappingId   MAPPING-ID85
target      _:ontop-bnode-18{ingredientevents_subject_id}/{ingredientevents_hadm_id}/{ingredientevents_stay_id}/{ingredientevents_caregiver_id}/{ingredientevents_starttime}/{ingredientevents_endtime}/{ingredientevents_storetime}/{ingredientevents_itemid}/{ingredientevents_amount}/{ingredientevents_amountuom}/{ingredientevents_rate}/{ingredientevents_rateuom}/{ingredientevents_orderid}/{ingredientevents_linkorderid}/{ingredientevents_statusdescription}/{ingredientevents_originalamount}/{ingredientevents_originalrate} <http://eg.com/ingredientevents#ref-subject_id> <http://eg.com/patients/subject_id={patients_subject_id}> . 
; target        _:ontop-bnode-18{ingredientevents_subject_id}/etc <http://eg.com/ingredientevents#ref-subject_id> <http://eg.com/patients/subject_id={patients_subject_id}> .
[...]

PR #866, opened 2025-05-14
Arity violation issue

ManuelaCorte:
This PR fixes issue #858.

The problem was that when building `AbstractOrNullFunctionSymbol`, terms that were simplified to either `true` or `false` constants were removed but the function arity was not updated, causing the function to expect more arguments than those actually present.


core/model/src/main/java/it/unibz/inf/ontop/model/term/functionsymbol/db/impl/AbstractOrNullFunctionSymbol.java
```diff
@@ -58,27 +58,6 @@ public boolean canBePostProcessed(ImmutableList<? extends ImmutableTerm> argumen
         return true;
     }
 
-    @Override
```
bcogrel, line 61: Please keep most of the logic in the class. Only the creation of function symbol (with a different arity) should be delegated to the subclasses.

justin2004:
thanks @ManuelaCorte !

PR #867, opened 2025-05-15
Recognize DuckDB catalogs

ManuelaCorte:
By default, DuckDB uses the database file name as the catalog name, or `memory` if no file is provided.

Previously, Ontop only supported tables in the default catalog, as it did not take the catalog name into account. This PR adds support for accessing tables in other catalogs by allowing users to specify the full table path (e.g., `catalog_name.schema_name.table_name`) in the mappings.


kontchakov (approved):

db/rdb/src/main/java/it/unibz/inf/ontop/dbschema/impl/DuckDBDBMetadataProvider.java
```diff
@@ -26,12 +26,14 @@ protected ResultSet getPrimaryKeysResultSet(String catalog, String schema, Strin
                 "generate_subscripts(constraint_column_names, 1) AS KEY_SEQ " +
                 "FROM duckdb_constraints " +
                 "WHERE " +
+                "database_name = ? AND " +
```
bcogrel, line 29: Just to double-check: is the database name always available? Does it work in simple settings where no remote database is attached?

ManuelaCorte: I'm not sure I'm interpreting the question correctly, but if there is a simple DuckDB instance specified in the properties file (e.g., `jdbc.url = jdbc:duckdb:path/to/database.db`) then the database name/catalog will be `database`. If instead the database is created in memory then the catalog name will be `memory`

bcogrel: Ok, thanks! So there is always a catalog.

bcogrel (approved):

Issue #458, opened 2021-08-23
allow multiple lines in target


majidaldo:
### Description

In the .obda file, the `source` field allows multiple lines but `taget` does not.

### Versions

Which version of Ontop are you using and in which setting (Protege, SPARQL endpoint, Cli, Java API, ...)
4.1.1 cli

ghxiao:
Hi @majidaldo, thanks for reporting this. 

We don't have a plan to support multiline targets In the OBDA mapping format in the near future. The reason is that our parser cannot be easily extended to support it. Allowing multiline targets requires redesigning the grammar.

Also, our Protege plugin reformats (e.g. adding proper spacing, shortening IRIs using prefixes) the input target from users and does not preserve the raw input.

majidaldo:
> Hi @majidaldo, thanks for reporting this.
> 
> We don't have a plan to support multiline targets In the OBDA mapping format in the near future. The reason is that our parser cannot be easily extended to support it. Allowing multiline targets requires redesigning the grammar.
> 
> Also, our Protege plugin reformats (e.g. adding proper spacing, shortening IRIs using prefixes) the input target from users and does not preserve the raw input.

i don't understand why you can't just pull out the `target` field first, remove the line breaks, then pass /that/ on to the parser as a quick fix. 

anyways, i think yaml for the obda file would have been easier. it makes it harder to review and make changes.

vemonet:
Multilines `target` should be built-in. Target can get quite large, it does not make sense to ask people to write it on one line. You don't need to rewrite your parser, just write a flattening function

You can use this python script in the meantime:

```python
import sys

def normalize_targets(obda_path: str, output_path: str):
    """Normalize the target lines in an OBDA file by merging multiple lines into a single line for each target."""
    with open(obda_path, "r", encoding="utf-8") as infile:
        lines = infile.readlines()
    result = []
    i = 0
    while i < len(lines):
        line = lines[i]
        stripped = line.strip()
        if stripped.startswith("target"):
            # Start of target
            target_line = line.rstrip("\n")
            i += 1
            # Collect subsequent lines until next mappingId, target, or ]]
            while i < len(lines):
                next_line = lines[i]
                next_stripped = next_line.strip()
                if next_stripped.startswith("mappingId") or next_stripped.startswith("target") or next_stripped.startswith("source") or next_stripped == "]]":
                    break
                target_line += " " + next_stripped
                i += 1
            result.append(target_line + "\n")
        else:
            result.append(line)
[...]

PR #874, opened 2025-06-05
Multi-line target queries in OBDA files

kontchakov:
Better handling of labels (`mappingId`, `target`, `source`) in OBDA files allows introduction of multi-line target queries for command-line tools (to address issue #458). 

Note, however, that the Protege plugin will not preserve line breaks in the target queries (in the same way as it does not preserve the order of mappings).


justin2004:
@kontchakov , thanks for adding this!

PR #875, opened 2025-06-11
Ontop support for TDengine

ManuelaCorte:
This pull request adds support for the time-series database TDengine in Ontop.

It is possible to create three different types of JDBC connections to TDengine:
- **Native connection**: Relies on TDengine’s native client library (libtaos.so for Linux, taos.dll for Windows, taos.dylib for MacOS).
- **REST connection**
- **WebSocket connection**: Not currently supported in Ontop

Since TDengine is a time-series database, some limitations apply:
- Tables must begin with a `TIMESTAMP` column acting as the primary key
- Some data types are not supported like `TIME` and `DECIMAL`.  Note that `DECIMAL` is only available starting from TDengine 3.3.6.0, and only via WebSocket connections.

Due to these constraints, the existing `lightweight-tests` suite isn't compatible. Instead, dedicated tests have been added (`CastTDEngineTest`, `ConstraintTDEngineTest`, `FunctionsTDEngineTest`) to verify the basic functionality of Ontop with TDengine.


bcogrel:
Thanks a lot Manuela!

PR #877, opened 2025-06-27
feature:  time-series database TDengine support WebSocket connection.

DuanKuanJun:
This pull request adds WebSocket connection for the time-series database TDengine in Ontop.

1.  connections only reserve WebSocket
TDengine supports three connection methods: Native, WebSocket, and Restful, but primarily promotes the WebSocket connection method. WebSocket has all the functionality of Restful and performs twice as well as Restful. The Native connection method has client compatibility issues, while WebSocket does not have this problem.
Therefore, only the WebSocket connection is retained here, and the other two have been removed.

2. add new class TDengineDBMetadataProvider to get database name and other compatible problems.


DuanKuanJun:
testIn2 and TestNotIn2  UT case with origin RESTFUL connection is failed, because float compare with double, precision not match, I fixed, please REVIEW & MERGE PR, Thanks.
@bcogrel    @ManuelaCorte

DuanKuanJun:
> Hi @DuanKuanJun, thanks for the PR!
> 
> Personally, I would keep the option to use the Native and Restful connections and I would specify in the documentation that, while the Native and Restful connections are possible, the recommended way to connect to TDengine is through the Websocket connection.
> 
> Also, if we keep Native and Restful connections it would make sense to run the tests against both the Restful and Websocket connection to make sure that the results are the same.

That's OK, keep Native and Restful connections but advise user connection with WebSocket in document. 
Am I modifying these on my branch or are you modifying them internally?

DuanKuanJun:
> Thanks a lot @DuanKuanJun for this PR!

Great, Thanks for REVIEW.

DuanKuanJun:
When was the approximate release time for Ontop 5.4.0 ? @bcogrel

PR #879, opened 2025-07-03
Upgrade the deployment process to Maven Central

bcogrel:
The former approach for deploying snapshots and releases is not valid anymore, as OSSRH service as terminated: 
https://central.sonatype.org/pages/ossrh-eol/

This PR upgrades the process to publish on Maven Central. It has been successfully tested for publishing the snapshots.


bcogrel:
Merging for now, we can revisit it later if needed.

PR #882, opened 2025-07-21
Case-insensitive substring matching

ManuelaCorte:
This PR introduces an additional optimization for the SPARQL `REGEX` function, simplifying `UCASE/LCASE` functions in the `REGEX` arguments when the case-insensitive flag is set. For example, the following expression:

```sparql
REGEX(LCASE(?x), "abc", "i")
```
will be simplified to:

```sparql
REGEX(?x, "abc", "i")
```
This optimization is particularly useful if the underlying database has indexes on the column used in the `REGEX` function, as the upper/lower case conversion can prevent the use of these indexes.

### DIFF
--- a/db/rdb/src/main/java/it/unibz/inf/ontop/model/term/functionsymbol/db/impl/PostgreSQLDBFunctionSymbolFactory.java
+++ b/db/rdb/src/main/java/it/unibz/inf/ontop/model/term/functionsymbol/db/impl/PostgreSQLDBFunctionSymbolFactory.java
@@ -395,9 +395,7 @@ public class PostgreSQLDBFunctionSymbolFactory extends AbstractSQLDBFunctionSymb
                     ImmutableTerm flagsTerm;
                     if (terms.get(2) instanceof Constant) {
                         Constant flagsConstant = (Constant) terms.get(2);
-                        flagsTerm = flagsConstant.getValue().contains("s")
-                                ? termFactory.getDBStringConstant(flagsConstant.getValue().replace("s", "n"))
-                                : termFactory.getDBStringConstant(flagsConstant.getValue());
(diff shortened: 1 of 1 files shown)


core/model/src/main/java/it/unibz/inf/ontop/model/term/functionsymbol/CaseSPARQLFunctionSymbol.java
```diff
@@ -0,0 +1,4 @@
+package it.unibz.inf.ontop.model.term.functionsymbol;
+
+public interface CaseSPARQLFunctionSymbol {
```
bcogrel, line 3: Could you please make it extend an existing interface?

bcogrel: I suggest renaming it into `CasingSPARQLFunctionSymbol`, to avoid the confusion with the SQL `CASE` expression.

db/rdb/src/main/java/it/unibz/inf/ontop/model/term/functionsymbol/db/impl/DuckDBDBFunctionSymbolFactory.java
```diff
@@ -162,6 +162,11 @@ public DBBooleanFunctionSymbol getDBRegexpMatches2() {
         return (DBBooleanFunctionSymbol) this.regexpLikeFunctionSymbol;
     }
 
+    @Override
+    public DBBooleanFunctionSymbol getDBRegexpMatches3() {
+        throw new UnsupportedOperationException("DuckDB Regex matches with three arguments is not supported by Ontop.");
```
bcogrel, line 167: This message is misleading: we don't know how to translate the Ontop expression into DuckDB dialect, not the other way around

bcogrel:
Thanks!

PR #884, opened 2025-07-23
DISCUSSION: user impersonation / bearer token authorization

justin2004:
if you have multiple users with varying access to the underlying data stores (trino, postgres, etc.) it can be necessary to allow ontop to impersonate each user running queries.

currently in a properties file if you do something like:
```
jdbc:trino://sometrino.server:8443/some-catalog/some-schema?SSL=true&accessToken=BEARER_TOKEN_HERE
```

it allows ontop to initialize BUT then ontop can't impersonate the user that runs the SPARQL query.

this PR allows you to set your properties like like the example but then if a query to ontop arrives with an HTTP `Authorization: bearer BEARER_TOKEN_HERE` header then it will remove the `accessToken` from the JDBC URL and it will append the `Authorization` header to the outgoing SQL query (thereby allowing ontop to run SQL as user X).

this approach is specific to trino with OAUTH authentication so i don't expect it to be merged as is but perhaps it could lead to a more general user impersonation solution.

### DIFF
--- a/client/endpoint-core/src/main/java/it/unibz/inf/ontop/endpoint/processor/AuthorizationContext.java
+++ b/client/endpoint-core/src/main/java/it/unibz/inf/ontop/endpoint/processor/AuthorizationContext.java
@@ -1,0 +1,52 @@
+package it.unibz.inf.ontop.endpoint.processor;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Thread-local context to track whether the current request has an Authorization header.
+ * This is used to coordinate JDBC URL modification between the SPARQL query processor
+ * and the JDBC connection creation layer.
+ */
+public class AuthorizationContext {
+    
+    private static final Logger logger = LoggerFactory.getLogger(AuthorizationContext.class);
+    
+    private static final ThreadLocal<String> AUTHORIZATION_TOKEN = new ThreadLocal<>();
+    
+    /**
+     * Sets the Authorization header value for the current request.
+     */
+    public static void setAuthorizationHeader(String authHeader) {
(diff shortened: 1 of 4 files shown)


bcogrel:
Hi @justin2004, thank you for opening this discussion on user impersonation with a concrete implementation. It is indeed an important topic and currently Ontop is missing the right abstractions to handle it.

Your PR handles the particular case of using the [JWT authentication](https://trino.io/docs/current/security/jwt.html) of Trino, where the JWT token is directly passed to Ontop as a bearer token in the HTTP Authorization header and Ontop directly passes it to Trino without doing any authentication on its side. So here the user query the Ontop SPARQL endpoint with the credentials for another service, Trino.

I foresee that a general solution for user impersonation should also support:
   - Having a middleware/API platform above the Ontop endpoints (that's what we do at Ontopic)
   - Letting Ontop be aware of the user identity to apply access control and/or adapt the SQL query according to the corresponding permissions (see https://github.com/ontop/ontop/pull/753)
   - Exchanging OAuth 2 tokens as the audiences of the Ontop service and the database may be different
   - Reuse JDBC connections for a given user
   - Passing the token to the data source according to its DB-specific conventions (no standard here)

Interestingly, this capability is about to arrive in the Postgres world, with [PG 18](https://www.postgresql.org/docs/18/auth-oauth.html). It is common with Databricks and Snowflake.

I think the `QueryContext` object should be used and extended to carry the token around with the query.

An alternative architecture would be to use Ontop as a query reformulation service and then have the client send the generated SQL query directly to Trino, if the latter is directly reachable by the client. However, at the moment, the generated SQL query doesn't perfectly match the SPARQL query as an extra post-processing step is needed. We are planning to add this capability next year.

giovannidegani:
How about adding support to that to the databricks connector ?

PR #898, opened 2025-08-29
Union graph (magic IRI)

justin2004:
in apache jena you can run queries like:
```sparql
PREFIX cco: <http://www.ontologyrepository.com/CommonCoreOntologies/>
PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
SELECT * WHERE {
  graph <urn:x-arq:UnionGraph> {
  ?s a cco:InformationBearingEntity .
  ?s cco:designated_by/cco:has_text_value ?name .
} 
}
LIMIT 101
```

which allow you to do property paths across named graphs (e.g. if names and InformationBearingEntities are in a different named graphs).

this PR adds that support to ontop.

`urn:x-arq:UnionGraph` described here:
https://jena.apache.org/documentation/tdb/datasets.html

### DIFF
--- a/binding/rdf4j/src/test/java/it/unibz/inf/ontop/rdf4j/repository/BasicGraphTest.java
+++ b/binding/rdf4j/src/test/java/it/unibz/inf/ontop/rdf4j/repository/BasicGraphTest.java
@@ -1,0 +1,91 @@
+package it.unibz.inf.ontop.rdf4j.repository;
+
+import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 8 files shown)


justin2004:
hey @ManuelaCorte , i think we can close this until [RDF4J supports quad construction](https://github.com/eclipse-rdf4j/rdf4j/discussions/5411).

PR #906, opened 2025-10-29
Refactoring IQTree creation and transformation

kontchakov:
This PR reviews and refactors a large number of classes where IQTrees are created and manipulated. The main purpose is to reduce code duplication and make code more transparent by abstracting away typical steps into new methods in `IQTreeTools` and related classes.

- `IQTreeTools` contains helper methods that wrap up short sequences of `IntermediateQueryFactory` calls into a single step (e.g., `createOptionalOrderByNode`, `createOptionalInnerJoinTree` etc.) These calls are also supported by `UnaryIQTreeBuilder`, which allows one to deal with optional unary nodes. This reduces the need for branching in the code that constructs instances of `IQTree`.
- `UnaryIQTreeTools.UnaryIQTreeDecomposition`, `NaryIQTreeTools.UnionDecomposition`, `NaryIQTreeTools.InnerJoinDecomposition` and `BinaryNonCommutativeIQTreeTools.LeftJoinDecomposition` are useful for decomposing instances of `IQTree` without two-step casts of the node and the subtree (using `static` imports for these decomposition classes is advisable). These classes also contain some utility methods, for example, for identifying co-occurring variables.
- The `IQTreeVisitor` hierarchy has been reshaped and simplified. Most visitors are now subclasses of `DefaultRecursiveIQTreeVisitingTransformer`, which processes the `IQTree` recursively. Also, almost all visitors now are turned into `private` inner classes to avoid excessively deep hierarchies with unclear behaviour. Thus, such visitors can be re-implemented with `instanceof` later (if needed to make code clearer). Note that visitors are typically not `static` so that they can access variables in the enclosing class. 
- `AbstractIQOptimizer` and `IQTreeVariableGeneratorTransformer` simplify many `IQTree` transformation step by making dependency on `VariableGenerator` clearer.
- `IQTreeVariableGeneratorTransformer` contains `static` methods for constructing more complex `IQTree` transformation steps in a compositional manner by using sequencing and iteration. `DefaultDelegatingIQTreeVariableGeneratorTransformer` can be used to avoid code duplication when composing individual `IQTreeVariableGeneratorTransformer`s and `IQTreeTransformer`s.
- `NormalizationContext` is introduced to implement all the `IQTree` normalisation steps in a uniform way. `NormalizationContext.State` represents a lightweight decomposition of an `IQTree`, which consists of a sequence of unary nodes, followed by the *focus node*(which is typically being normalised). This class also has methods that help implementing fixpoint computations. Note also that dedicated `Normalizer` classes have split form the `QueryNode` classes.  
- `AbstractTermTransformer` contains a single abstract method `replaceFunctionSymbol` that either returns a replacement or the empty `Optional` to indicate that the functional term remains unchanged.
- `DownPropagation` is introduced as an umbrella for the following three operations in `IQTree`: `applyFreshRenaming`, `propagateDownConstraint` and `applyDescendingSubstitution`; the implementation of these operations has been more consistent. Instances of `DownPropagation` are created by `IQTreeTools`. Note that `applyDescendingSubstitutionWithoutOptimization` has been eliminated as its uses were not essential (most occurrences were in fact performing renaming).


binding/rdf4j/src/test/java/it/unibz/inf/ontop/rdf4j/repository/DestinationFullSQLReformulationTest.java
```diff
@@ -73,14 +73,16 @@ public void testQuery() {
 
     /**
      * SPARQL query rejected because it is not strongly typed
+     * It was not normalized fully - 3 consecutive CONSTRUCT
      */
-    @Test(expected = QueryEvaluationException.class)
+    @Test//(expected = QueryEvaluationException.class)
```
bcogrel: The test passes as it takes 10 triples from the Tbox about classes. These are all typed the same way. Raising the limit to 100 000 triggers the exception due to multi-typing

bcogrel (approved):

bcogrel:
Thanks a lot Roman for this important refactoring!

PR #911, opened 2025-11-17
Common Table Expression in SQL serialization

kontchakov:
Removes code duplication in various implementations of `SelectFromWhereSerializer` and introduces `WITH` clauses (common table expressions, CTEs) for the serialization of Ontop black box views (complex SQL queries in mappings). 

Property `ontop.useCTEsForBlackBoxViewsIfSupported` (`boolean`) switches the feature off.

Property `ontop.commonTableExpressionsPrefix` (`String`) specifies the prefix of produced CTE identifiers (`ontopCTE` is the default).


db/rdb/src/main/java/it/unibz/inf/ontop/injection/impl/OntopSQLCoreSettingsImpl.java
```diff
@@ -43,6 +45,10 @@ protected OntopSQLCoreSettingsImpl(Properties userProperties) {
 
         jdbcUrl = getRequiredProperty(OntopSQLCoreSettings.JDBC_URL);
         jdbcDriver = getRequiredProperty(OntopSQLCoreSettings.JDBC_DRIVER);
+        useCTEs = getBoolean(ONTOP_USE_COMMON_TABLE_EXPRESSIONS_FOR_BLACK_BOX_VIEWS_IF_SUPPORTED)
```
bcogrel: Please put the default values into the `sql-default.properties` file instead

PR #926, opened 2026-01-16
Update Java and dependency versions to resolve CVEs

zadeluca:
- Java 11 -> 17
- Spring Boot 2.7.9 -> 3.5.9


zadeluca:
Hello, I work with @justin2004. We recently made a push to reduce our CVE counts, and I was hoping you would consider this PR to upgrade Java and Spring Boot versions. It appears to be working well for us, but could probably benefit from additional testing. Please let me know if I missed anything, thanks!

PR #932, opened 2026-03-18
Circular dependency involving the OptimizationSingletons fixed

ManuelaCorte:
In order to solve a circular dependency involving the `OptimizationSingletons` class, this PR refactors some classes to receive the required dependencies (`CoreSingletons` and `DefinitionPushDownTransformer`) via constructor injection instead of accessing them directly from the `OptimizationSingletons` class.

This circular dependency is causing problems to downstream applications. For Ontop itself, Guice managed to handle it automatically.

### DIFF
--- a/core/optimization/src/main/java/it/unibz/inf/ontop/injection/OptimizationSingletons.java
+++ b/core/optimization/src/main/java/it/unibz/inf/ontop/injection/OptimizationSingletons.java
@@ -17,17 +17,11 @@ public interface OptimizationSingletons {
 
     CoreSingletons getCoreSingletons();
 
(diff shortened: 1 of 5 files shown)


kontchakov:
Perhaps, OptimizationSingletons can be slimmed down too - most of its methods have no use (and classes like DefinitionPushDownTransformer are injected anyway). @bcogrel, what do you think?

bcogrel (approved):

bcogrel:
> Perhaps, OptimizationSingletons can be slimmed down too - most of its methods have no use (and classes like DefinitionPushDownTransformer are injected anyway). @bcogrel, what do you think?

Yes, it makes sense, thanks. We will clean this interface before merging.

PR #933, opened 2026-03-21
SPARQL to SQL reformulation for regular or native consumption can now be decided at runtime

bcogrel:
This PR refactors the full SPARQL-to-SQL reformulation so it can be activated at runtime through a parameter instead of being configured at init time. The configuration key `ontop.reformulateToFullNativeQuery` is removed.

The full reformulation now preserves the variable ordering from the SPARQL query in the generated SQL query. To achieve this, `NativeNode.getVariables()` has been refactored: it now uses a regular `ImmutableSet` instead of an `ImmutableSortedSet`. The `ImmutableSet` preserves the order.

`QueryCache` now takes `forNativeConsumption` into account.

The parameter `forNativeConsumption=true` is now supported by the reformulation HTTP endpoint. It triggers the full reformulation.


db/rdb/src/main/java/it/unibz/inf/ontop/generation/algebra/impl/IQTree2SelectFromWhereConverterImpl.java
```diff
@@ -222,15 +220,15 @@ public SQLExpression transformInnerJoin(NaryIQTree tree, InnerJoinNode rootNode,
...
                 ImmutableList<SQLExpression> subExpressions = NaryIQTreeTools.transformChildren(
                         tree.getChildren(),
                         c -> convert(c, signature));
                 return sqlAlgebraFactory.createSQLUnionExpression(subExpressions, unionNode.getVariables());
             }
 
-            private ImmutableSortedSet<Variable> getSignature(IQTree tree) {
-                return ImmutableSortedSet.copyOf(tree.getVariables());
+            private ImmutableSet<Variable> getSignature(IQTree tree) {
+                return ImmutableSet.copyOf(tree.getVariables());
```
kontchakov: Do we really need a copy here? IQTree.getVariables gives us an ImmutableSet anyway.

db/rdb/src/main/java/it/unibz/inf/ontop/generation/algebra/SelectFromWhereWithModifiers.java
```diff
@@ -18,7 +17,10 @@
  */
 public interface SelectFromWhereWithModifiers extends SQLExpression {
 
-    ImmutableSortedSet<Variable> getProjectedVariables();
+    /**
+     * The order in the ImmutableSet matters
```
kontchakov: is the insertion order?

engine/reformulation/core/src/main/java/it/unibz/inf/ontop/answering/reformulation/impl/GuiceBasedQueryCache.java
```diff
@@ -11,15 +10,42 @@
...
+            this.forNativeConsumption = forNativeConsumption;
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (!(o instanceof CacheKey)) return false;
+            CacheKey other = (CacheKey) o;
+            return forNativeConsumption == other.forNativeConsumption
+                    && Objects.equals(inputQuery, other.inputQuery)
```
kontchakov: If these are non-nullable, then .equals does the job.

engine/reformulation/sql/src/main/java/it/unibz/inf/ontop/answering/reformulation/generation/impl/SQLGeneratorImpl.java
```diff
@@ -60,7 +81,8 @@ private SQLGeneratorImpl(@Assisted DBParameters dbParameters,
                              IQTree2NativeNodeGenerator defaultIQTree2NativeNodeGenerator,
                              DialectExtraNormalizer extraNormalizer, BooleanExpressionPushDownTransformer pushDownTransformer,
                              EmptyRowsValuesNodeTransformer valuesNodeTransformer,
-                             OntopReformulationSQLSettings settings, ExplicitEqualityTransformer equalityTransformer, IQTreeTools iqTreeTools)
+                             OntopReformulationSQLSettings settings, ExplicitEqualityTransformer equalityTransformer, IQTreeTools iqTreeTools,
```
kontchakov: CoreSingletons instead?

kontchakov (approved):
Looks good! Just a couple of little suggestions in the comments.

db/rdb/src/main/java/it/unibz/inf/ontop/generation/algebra/SelectFromWhereWithModifiers.java
bcogrel: Yes, updated

bcogrel:
> Looks good! Just a couple of little suggestions in the comments.

Thanks Roman!

PR #938, opened 2026-06-04
Cannot create a lens with Athena

ManuelaCorte:
Fixes relation resolution for AWS Athena so that the same table can be referenced through any of the catalog/schema/table case variants Athena accepts at the SQL level (e.g. `awsdatacatalog.books.books`, `"AWSDATACATALOG"."books"."books"`, `"books"."books"`, …).

## Root cause

The Simba Athena JDBC driver's `DatabaseMetaData.getColumns(catalog, schema, table, …)` is case-sensitive on the catalog argument, even though Athena SQL itself is case-insensitive at the identifier-resolution level.

## Fix

Override `AthenaDBMetadataProvider.getColumnsResultSet` to bypass the Simba JDBC implementation and query `INFORMATION_SCHEMA.COLUMNS` directly.

This trades one driver-specific quirk for a SQL one: `INFORMATION_SCHEMA` stores all identifiers lowercased, while the canonical `RelationID` we resolve may carry a different case, either because the user wrote it that way in the mapping, or because the catalog was padded from the JDBC URL's `Catalog=…` parameter. Returning the row's stored values would make the row's identifier disagree with the canonical id (consistency check in `extractRelationID`). Instead of selecting from the table, we echo the bound parameters back for `TABLE_CAT`/`TABLE_SCHEM`/`TABLE_NAME`.

The remaining JDBC-contract columns are populated as follows:

| Column                            | Source                | Notes |
|---                                |---                    |---|
| `COLUMN_NAME`                     | `column_name`         |  |
| `TYPE_NAME`                       | `data_type`           | already carries parameters with precision (`varchar(100)`, `decimal(38,2)`) |
| `DATA_TYPE`                       | `Types.OTHER` (1111)  | placeholder. Only used to append precision which is already present in `TYPE_NAME` |
| `COLUMN_SIZE`, `DECIMAL_DIGITS`   | `0`                   | placeholders, unused |
| `NULLABLE`                        | `CASE WHEN is_nullable = 'YES' THEN 1 ELSE 0 END` | maps Athena's `'YES'`/`'NO'` to the JDBC int contract (`columnNullable=1`, `columnNoNulls=0`) |

### DIFF
--- a/db/rdb/src/main/java/it/unibz/inf/ontop/dbschema/impl/AthenaDBMetadataProvider.java
+++ b/db/rdb/src/main/java/it/unibz/inf/ontop/dbschema/impl/AthenaDBMetadataProvider.java
@@ -5,11 +5,14 @@ import com.google.inject.assistedinject.AssistedInject;
 import it.unibz.inf.ontop.dbschema.RelationID;
 import it.unibz.inf.ontop.exception.MetadataExtractionException;
 import it.unibz.inf.ontop.injection.CoreSingletons;
+import it.unibz.inf.ontop.utils.ImmutableCollectors;
 
 import java.sql.Connection;
+import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.sql.Types;
 
 public class AthenaDBMetadataProvider extends TrinoDBMetadataProvider {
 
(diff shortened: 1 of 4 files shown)


bcogrel (approved):

bcogrel:
Thanks Manuela!

PR #941, opened 2026-06-18
Replaced PreparedStatement for Athena

ManuelaCorte:
The PreparedStatement used to extract metadata for Athena was replaced with a normal Statement as prepared statements require additional IAM permissions in Athena (https://docs.aws.amazon.com/athena/latest/ug/security-iam-athena-prepared-statements.html)


bcogrel:
Thanks Manuela!

PR #942, opened 2026-06-18
Bugfix/out of memory

kontchakov:
Fixes the out of memory and stack overflow bugs caused by checking possible solution mapping compatibility conditions with large VALUES nodes.


bcogrel (approved):

bcogrel:
Nice one, thanks Roman!

PR #945, opened 2026-07-13
Rejected SQL translation for  PostgreSQL "text #>> unknown"

ManuelaCorte:
On PostgreSQL, serializing a timestamp to its lexical form is translated as `TO_JSON(<ts>)#>>'{}'`. When the serialized timestamp is concatenated into a larger string, PostgreSQL rejects the query with:

```
ERROR: operator does not exist: text #>> unknown
```

This is due to operator precedence: `#>>` shares its precedence level with `||` and is left-associative. Since `TO_JSON(...)#>>'{}'` was emitted without wrapping parentheses, an expression such as:

```sql
('published on ' || TO_JSON(v1."publication_date")#>>'{}')
```

is parsed left-to-right as `('published on ' || TO_JSON(v1."publication_date")) #>> '{}'`, so `#>>` is applied to a `text` value (the concatenation) instead of to `TO_JSON(...)`.

The fix was to simply wrap the serialization in parentheses:

```sql
('published on ' || (TO_JSON(v1."publication_date")#>>'{}'))
```


bcogrel:
Thanks a lot Manuela!

PR #950, opened 2026-09-21
Dependency updates for security fixes

ManuelaCorte:
This PR updates the dependencies flagged by Dependabot, plus two transitive ones it cannot see, to versions that contain the corresponding security fixes. Ontop still has to run on Java 11, so Spring Boot stays on 2.7.18, and every update below is limited to what remains compatible with it. Updates stay as minimal as possible wherever the fix allows it.

### Changes

| Dependency | Before | After | Notes |
|---|---|---|---|
| Jackson | 2.18.6 | 2.18.11 | A `jackson-bom` import is added so that Jackson modules arriving only transitively (`jackson-datatype-jsr310`, `jackson-module-parameter-names`, both via Spring Boot) follow the same version. |
| Tomcat embed | 9.0.117 | 9.0.122 | `tomcat-embed-el` and `tomcat-embed-websocket` are now pinned to the same version as `tomcat-embed-core`. |
| Thymeleaf | 3.0.15.RELEASE | 3.1.5.RELEASE | The 3.0 line has no patched release. `thymeleaf-spring5` is pinned alongside so both artifacts move together. The endpoint's only template uses `th:text` on a String attribute and touches nothing that changed in 3.1. |
| log4j2 | 2.19.0 | 2.25.5 | No log4j2 artifact is shipped; it only reached the `lightweight-tests` classpath through `hive-jdbc`. The bump clears the alerts raised on the managed versions. |
| Spring Framework | 5.3.31 (Spring Boot 2.7.18 default) | 5.3.39 | Not visible to Dependabot. Applied by importing `spring-framework-bom`; 5.3.39 is the last open-source 5.3.x release and is compatible with Spring Boot 2.7. |
| Commons BeanUtils | 1.9.4 (via `opencsv` 5.7.1) | 1.11.0 | Not visible to Dependabot. Transitive dependency of `rdf4j-queryresultio-text`. |
| `spring-boot-maven-plugin` | 2.7.9 | 2.7.18 | Aligned with `spring-boot.version`. Only used by the `build-info` goal in the release profile. |
| `lightweight-tests`: `hive-jdbc` exclusions | – | log4j2 excluded | Hive's JDBC client logs through SLF4J and Logback is the test binding, so the log4j2 stack it pulls in is unused. Also, log4j 2.25 conflicted at compile time with the signed Simba Athena driver jar. |

### Notes

- The JDBC drivers in `lightweight-tests` and `docker-tests` are only used for testing. Since changing them would require extensive testing and they are never shipped, they were not updated. The alerts they have can be dismissed as test-only.

- The remaining Spring Boot and Logback alerts have no fix compatible with Java 11: Spring Boot 2.7.18 is the last 2.7 release and Spring Boot 3 requires Java 17, while Logback 1.3 is incompatible with Spring Boot 2.7's logging configurator.


bcogrel (commented):
Thanks. Does it work without the two `BOM` dependencies? 
I would keep changes minimal

ManuelaCorte:
> Thanks. Does it work without the two `BOM` dependencies? I would keep changes minimal

Yes everything would work without the BOM dependencies. 

`jackson-bom` only aligns transitive dependencies with no currently known vulnerability so it can be dropped.

However, `spring-framework-bom` lifts Spring Framework from 5.3.31 (what Boot 2.7.18 brings) to 5.3.39 and fixes fives CVEs that Dependabot does not report because Spring Framework is only a transitive dependency:
- https://spring.io/security/cve-2024-38809 (high)
- https://spring.io/security/cve-2024-22262 (high)
- https://spring.io/security/cve-2024-38808 (high)
- https://spring.io/security/cve-2024-22243 (medium)
- https://spring.io/security/cve-2024-22259 (medium)

Let me know if you want me to remove it anyway

